PR #11, opened 2023-09-26
First version of VTX digitisation for IDEA vertex detector

armin-ilg:
Hi @brieucf

Here is a first version of the digitisaton for the vertex detector/silicon detectors. 
For the moment, it only applies a Gaussian smearing to the hits. I will refine this more over time.

One issue I still have is that I cannot do the digitisation of the various vertex subdetectors simultaneously, when uncommenting all VTXdigitizers ([here](https://github.com/armin-ilg/k4RecTracker/blob/aa306b2786a4a031876c18f860c15df2e15209b2/VTXdigi/test/runVTXdigitizer.py#L216)).

I will also make sure that it works properly with the CLD vertex detector.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,6 @@ MAKEFLAGS += --no-print-directory
 make:
 	@ mkdir -p build install ; \
 	cd build ; \
-	source /cvmfs/sw-nightlies.hsf.org/key4hep/setup.sh ; \
 	cmake .. -DCMAKE_INSTALL_PREFIX=../install ; \
 	make install -j4 ; \
 	cd .. ; \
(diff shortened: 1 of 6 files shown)


BrieucF:
> One issue I still have is that I cannot do the digitisation of the various vertex subdetectors simultaneously, when uncommenting all VTXdigitizers ([here](https://github.com/armin-ilg/k4RecTracker/blob/aa306b2786a4a031876c18f860c15df2e15209b2/VTXdigi/test/runVTXdigitizer.py#L216)).

Each digitizer should have a different name: https://github.com/armin-ilg/k4RecTracker/blob/aa306b2786a4a031876c18f860c15df2e15209b2/VTXdigi/test/runVTXdigitizer.py#L170C32-L170C44 ?

armin-ilg:
Hi @BrieucF 

It should be okay now, forcing the hits onto the surfaces works. k4RecTracker can be used to digitise the IDEA vertex or Silicon Wrapper or any of the CLD tracking layers.
Shall we merge it? Do you know what is wrong with all the tests? Running 
`k4run test/runVTXdigitizer.py`
works for me locally

BrieucF:
Hi Armin, thanks! Merging this for now, we can do the modifications in a second PR:

- migrate configuration files to ddsim workflow (using the ddsim output and removing the k4SimGeant4 related steps)
- use a thread safe random number generator (align with k4Reco)
- have association between tracker hits and sim tracker hits out of the digitizer: https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L457

PR #12, opened 2023-11-16
added the cluster counting algorithm

welmeten:


jmcarcell:
This is now quite outdated and it's going to be three years that it has been opened. Let me know if anyone is still interested.

PR #21, opened 2024-04-26
Transformer to create tracks from gen particles

BrieucF:
BEGINRELEASENOTES
- Add a transformer to create Tracks out of generated particles
- Add a consumer that plots the distance between simTrackerHits and a Track 
ENDRELEASENOTES

### DIFF
--- a/Tracking/components/PlotTrackHitResiduals.cpp
+++ b/Tracking/components/PlotTrackHitResiduals.cpp
@@ -1,0 +1,97 @@
+// Gaudi
+#include "Gaudi/Property.h"
+#include "GaudiAlg/Consumer.h"
+#include "Gaudi/Accumulators/RootHistogram.h"
+#include "Gaudi/Histograming/Sink/Utils.h"
+
+// edm4hep
+#include "edm4hep/MCParticleCollection.h"
+#include "edm4hep/TrackCollection.h"
+#include "edm4hep/MCRecoTrackParticleAssociationCollection.h"
+#include "edm4hep/SimTrackerHitCollection.h"
+
+// marlin
+#include <marlinutil/HelixClass_double.h>
+
+// ROOT
+#include "TH1D.h"
+
+// Define BaseClass_t
+#include "k4FWCore/BaseClass.h"
+
+#include <string>
+
+/** @class PlotTrackHitDistances
+ *
+ *  Gaudi consumer that generates a residual distribution (mm) by comparing the helix from Track AtIP and simHit position.
+ *  This is intended to be used on tracks produced from gen particles i.e. which do not have real hits attached to them.
+ *
+ *  @author Brieuc Francois
+ */
+
+struct PlotTrackHitDistances final
+  : Gaudi::Functional::Consumer<void(const edm4hep::SimTrackerHitCollection&, const edm4hep::MCRecoTrackParticleAssociationCollection&), BaseClass_t> {
+  PlotTrackHitDistances(const std::string& name, ISvcLocator* svcLoc)
+      : Consumer(
+            name, svcLoc,
+            {
+            KeyValue("InputSimTrackerHits", "DCHCollection"),
+            KeyValue("InputTracksFromGenParticlesAssociation", "TracksFromGenParticlesAssociation"),
+            }) {}
+
+  void operator()(const edm4hep::SimTrackerHitCollection& simTrackerHits, const edm4hep::MCRecoTrackParticleAssociationCollection& trackParticleAssociations) const override {
+
+    for (const auto& trackParticleAssociation : trackParticleAssociations) {
+      auto genParticle = trackParticleAssociation.getSim();
+      auto track = trackParticleAssociation.getRec();
+      edm4hep::TrackState trackStateAtIP;
+      bool found_trackStateAtIP = false;
+      for (const auto& trackState : track.getTrackStates()) {
+        if (trackState.location == edm4hep::TrackState::AtIP) {
+          trackStateAtIP = trackState;
+          found_trackStateAtIP = true;
+        }
+      }
+      if (!found_trackStateAtIP)
+        throw std::runtime_error("No track state defined AtIP, exiting!");
+
+      // Build an helix out of the trackState
+      auto helixFromTrack = HelixClass_double();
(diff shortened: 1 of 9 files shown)


Tracking/components/PlotTrackHitResiduals.cpp
```diff
@@ -0,0 +1,97 @@
...
+	StatusCode finalize() override {
+			TFile file("track_hit_distances.root", "RECREATE");
+			std::string name = "";
+			// Name that will appear in the stats table
+			std::string histName = "Distances";
+			nlohmann::json jH {m_residualHist};
+			auto [histo, dir] = Gaudi::Histograming::Sink::jsonToRootHistogram<Gaudi::Histograming::Sink::Traits<false, TH1D, 1>>(name, histName, jH[0]);
+      // Add overflow bin to the last bin
+      int lastBinIndex = histo.GetNbinsX();
+      histo.SetBinContent(lastBinIndex, histo.GetBinContent(lastBinIndex) + histo.GetBinContent(lastBinIndex + 1));
```
jmcarcell, line 88: Why this? You can easily show how many entries you get in the overflow and underflow bins when plotting. With this you also change the values of the mean and Std. Dev. that will be displayed and it's impossible to distinguish a histogram with many entries in the last bin from a histogram with many overflow entries.

jmcarcell (commented):
Some comments. I didn't run it, but it looks fine

Tracking/test/runTracksFromGenParticles.py
```diff
@@ -0,0 +1,52 @@
+from Configurables import ApplicationMgr
+from Configurables import EventCounter
+from Gaudi.Configuration import INFO, WARNING, DEBUG
+import os
+
+if not os.path.isfile("ddsim_output_edm4hep.root"):
+    os.system("ddsim --enableGun --gun.distribution uniform --gun.energy '10*GeV' --gun.particle e- --numberOfEvents 100 --outputFile ddsim_output_edm4hep.root --random.enableEventSeed --random.seed 42 --compactFile $K4GEO/FCCee/IDEA/compact/IDEA_o1_v02/IDEA_o1_v02.xml")
```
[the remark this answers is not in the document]
BrieucF: Hi Juan, thanks for the comments! Not sure I understand this one though. This line is just for convenience (and for tests). So that I can run/test the genTracking just calling one script

Tracking/components/PlotTrackHitResiduals.cpp
BrieucF: This is to visualize, by default the amount of hits that are far from the track. If you don't like I can remove it, not a big deal

jmcarcell (commented):
A couple more comments:
- In the steering file it would be good to get rid of `PodioInput` and `PodioOutput` and use `IOSvc` like in https://github.com/key4hep/k4FWCore/blob/971ae2da7efec5a167b4ab6ecf365c0e27ca41ca/test/k4FWCoreTest/options/ExampleFunctionalFile.py#L28. It should work with `PodioInput` and `PodioOutput` but if you ever want to run multithreaded it won't
- For histogramming I think that works fine, but since a few days ago there is an example in k4FWCore: https://github.com/key4hep/k4FWCore/blob/main/test/k4FWCoreTest/src/components/ExampleFunctionalTransformerHist.cpp and https://github.com/key4hep/k4FWCore/blob/main/test/k4FWCoreTest/options/ExampleFunctionalTransformerHist.py where in the cpp file you only need to fill the histogram and the exporting gets done by the `RootHistSvc`.

BrieucF:
> For histogramming I think that works fine, but since a few days ago there is an example in k4FWCore: https://github.com/key4hep/k4FWCore/blob/971ae2da7efec5a167b4ab6ecf365c0e27ca41ca/test/k4FWCoreTest/src/components/ExampleFunctionalTransformerHist.cpp and https://github.com/key4hep/k4FWCore/blob/971ae2da7efec5a167b4ab6ecf365c0e27ca41ca/test/k4FWCoreTest/options/ExampleFunctionalTransformerHist.py where in the cpp file you only need to fill the histogram and the exporting gets done by the RootHistSvc.

Is this really wanted/important? I need to access the TH1 to add the overflow bin

Tracking/components/PlotTrackHitResiduals.cpp
jmcarcell: But this means that now the plot has to be always accompanied with information on how it was made because the last bin that is displayed means something different from all the other bins. And also using some statistical quantity that depends on the bin content will be wrong. There is a way of showing the underflow and overflow counts in the statistics box that doesn't change, for example in python:
``` python
import ROOT
import numpy as np

hist = ROOT.TH1F("hist", "Sample Histogram", 100, -1, 1)

data = np.random.normal(0, 1, 10000)
for value in data:
    hist.Fill(value)

canvas = ROOT.TCanvas("canvas", "Histogram Canvas", 800, 600)
hist.Draw()

ROOT.gStyle.SetOptStat(1111111)
hist.SetStats(1)

canvas.Update()
canvas.Draw()

input()
```
and then the last bin doesn't have to be touched and is automatically computed by ROOT histograms.

BrieucF:
> For histogramming I think that works fine, but since a few days ago there is an example in k4FWCore: https://github.com/key4hep/k4FWCore/blob/971ae2da7efec5a167b4ab6ecf365c0e27ca41ca/test/k4FWCoreTest/src/components/ExampleFunctionalTransformerHist.cpp and https://github.com/key4hep/k4FWCore/blob/971ae2da7efec5a167b4ab6ecf365c0e27ca41ca/test/k4FWCoreTest/options/ExampleFunctionalTransformerHist.py where in the cpp file you only need to fill the histogram and the exporting gets done by the RootHistSvc.

Done

Tracking/components/PlotTrackHitResiduals.cpp
BrieucF: Fair enough

BrieucF:
@jmcarcell the test related to this PR passes (I removed Event Counter). Shall we handle the other tests in another PR?

jmcarcell:
In general a different PR is better because then people can look back and understand why the change was done, otherwise they will find this PR... 
Another comment is that Associations will have to be updated to Links but this is probably better done at once for the whole repository.

jmcarcell (approved):

PR #24, opened 2024-07-25
[MUONDigi] Adding a digitizer of IDEA Muon system.

mahmoudali2:
BEGINRELEASENOTES

Defining a first draft for the muon system digitizer, which can do the following tasks:
- Smearing the simHits in 2 Dimensions in the local coordinates of the chamber and applying the changes to the global coordinates of the Hits.
- Emulating the efficiency of the chamber.
- The generated root file will have branches to validate the smearing (Differences between DigiHits and SimHits in global coordinates).

ENDRELEASENOTES


MUONdigi/dataFormatExtension/muonSystemHit.yaml
```diff
@@ -0,0 +1,21 @@
...
+options:
+  # should getters / setters be prefixed with get / set?
+  getSyntax: True
+  # should POD members be exposed with getters/setters in classes that have them as members?
+  exposePODMembers: False
+  includeSubfolder: True
+
+datatypes:
+
+  extension::MCRecoMuonSystemDigiAssociation:
```
BrieucF, line 12: Why not using `MCRecoTrackerAssociation` https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L634 ?

MUONdigi/include/MUONsimpleDigitizer.h
```diff
@@ -0,0 +1,107 @@
...
+  // y resolution in mm
+  FloatProperty m_y_resolution{this, "yResolution", 1.0, "Spatial resolution in the y direction [mm]"};
+
+  // z resolution in mm
+  FloatProperty m_z_resolution{this, "zResolution", 1.0, "Spatial resolution in the z direction [mm]"};
+
+  // Detector efficiency
+  FloatProperty m_efficiency{this, "efficiency", 0.95, "Detector efficiency"};
+
+  // Declaration of validation distribution
```
BrieucF, line 92: Can you add a flag, set to false by default, to produce these collections? See e.g. https://github.com/key4hep/k4RecTracker/blob/master/DCHdigi/include/DCHsimpleDigitizerExtendedEdm.h#L78 (the branch will unfortunately still be there I am afraid)

MUONdigi/include/MUONsimpleDigitizer.h
```diff
@@ -0,0 +1,107 @@
...
+  DataHandle<podio::UserDataCollection<double>> m_simDigiDifferenceX{"simDigiDifferenceX", Gaudi::DataHandle::Writer, this}; // mm
+  DataHandle<podio::UserDataCollection<double>> m_simDigiDifferenceY{"simDigiDifferenceY", Gaudi::DataHandle::Writer, this}; // mm
+  DataHandle<podio::UserDataCollection<double>> m_simDigiDifferenceZ{"simDigiDifferenceZ", Gaudi::DataHandle::Writer, this}; // mm
+  
+  // Random Number Service
+  IRndmGenSvc* m_randSvc;
```
BrieucF, line 98: @jmcarcell let us know if a consensus was reached on which random number generator we should use

MUONdigi/include/MUONsimpleDigitizer.h
```diff
@@ -0,0 +1,107 @@
...
+  // Volume manager to get the physical cell sensitive volume
+  dd4hep::VolumeManager m_volman;
+
+  // x position resolution in mm
+  FloatProperty m_x_resolution{this, "xResolution", 1.0,
+                               "Spatial resolution in the x direction [mm]"};
+  // y resolution in mm
+  FloatProperty m_y_resolution{this, "yResolution", 1.0, "Spatial resolution in the y direction [mm]"};
+
+  // z resolution in mm
```
BrieucF, line 86: I don't understand why we need 3 numbers here. Since you do the smearing in local coordinates (as it should), 2 numbers should be sufficient no? OR at least we should put ine by default to 0, otherwise we really smear in the 3 dimensions.

BrieucF (commented):
Very nice Mahmoud! Here are already a few questions/comments

MUONdigi/dataFormatExtension/muonSystemHit.yaml
mahmoudali2: Can it be used for linking between digi and sim? I fount it relates between reco and sim.. 
On another hand I'd like to see if there is an association also to store the validation data (e.g: simDigipPositionDifference)

MUONdigi/src/MUONsimpleDigitizer.cpp
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    auto output_digi_hit = output_digi_hits->create();
+    // smear the hit position:
+    // retrieve the cell detElement
+    dd4hep::DDSegmentation::CellID cellID         = input_sim_hit.getCellID();
+    debug() << "Digitisation of " << m_readoutName << ", cellID: " << cellID << endmsg;
+    //  auto                           cellDetElement = m_volman.lookupDetElement(cellID);
+
+    const auto& stripsTransformMatrix = m_volman.lookupVolumePlacement(cellID).matrix();
```
atolosadelgado, line 90: maybe using the segmentation position and cellID functions [[link](https://dd4hep.web.cern.ch/dd4hep/reference/classdd4hep_1_1Segmentation.html#ab894adbb604c9786d2bd801a6c5fac66)] give the same functionality without the overhead of the volume manager?

atolosadelgado (commented):
Hi @mahmoudali2 

thank you for this PR. I made few comments, but we can talk offline further if you want :)

davidlange6:
should this be closed as being superceded by #84?

mahmoudali2:
> should this be closed as being superceded by #84?

Yes, I'm going to close it

mahmoudali2:
This has be replaced by #84

PR #25, opened 2024-07-27
Do not use GaudiAlg

jmcarcell:
These are the changes needed not to use GaudiAlg, which will be removed (maybe soon) in Gaudi.

BEGINRELEASENOTES
- Change headers and add EventContext in algorithms not to use GaudiAlg
- Replace `GaudiTool` with `AlgTool`

ENDRELEASENOTES


BrieucF:
Thanks!

PR #27, opened 2024-08-05
DCH v2 digitizer, which smears the position and adds cluster information

atolosadelgado:
BEGINRELEASENOTES
- New digitizer for DCH v2 (based on twisted tubes). It calculates cluster information and smears the position along and perpendicular the wire separately. New custom EDM4hep data extension was added to include these data.

ENDRELEASENOTES

Hi, 

Please find attached the drift chamber (DCH) digitizer, based on Walaa's code for the cluster number and size calculations. In addition, I added the position smearing based on analytical calculations (I was stuck here, but thanks to Andre we found the problem).

The algorithm works as expected, but I can imagine there is room for improvement regarding naming, coding style in general, so please let me know if I can improve it somehow.

I have been working on a stand alone repository, in case you want to try out: https://github.com/atolosadelgado/DCH_detector

Alvaro.


BrieucF:
Hi Alvaro, excellent! Can you add a test and add a link between sim and digi hits? See e.g. https://github.com/key4hep/k4RecTracker/pull/28/files

DCHdigi/dataFormatExtension/driftChamberHit.yaml
```diff
@@ -33,6 +34,24 @@ datatypes:
...
+    Author: "A. Tolosa-Delgado, B. Francois, CERN"
+    Members:
+      - uint64_t cellID               // ID of the sensor that created this hit
+      - int32_t type                  // type of the raw data hit
+      - int32_t quality               // quality bit flag of the hit
+      - float time                    // time of the hit [ns]
+      - float eDep                    // energy deposited on the hit [GeV]
+      - float eDepError               // error measured on eDep [GeV]
+      - edm4hep::Vector3d position    // point on the sensitive wire (SW) which is closest to the simhit [mm]
+      - edm4hep::Vector3d position_L  // point on the left to the sensitive wire (SW) [mm]
```
BrieucF, line 48: This is redundant with the other members, I was actually proposing to add a function in the DD4hep dch extensions to retrieve position_L and position_R from cellID and distanceToWire but I am not even sure it is needed, let's see in the future. Can you remove those two members?

atolosadelgado: I was thinking that kind of functionality is only useful for reconstruction and subsequent steps... maybe an alternative can be placing the functionality in a dedicated header-only file, which will be included by this algorithm, and since it will be installed it can be included by other projects. Actually, we can put there not only that function, but other functions such as the ones calculating the distance to the wire, or the functions that calculate the position of the wire based on the data extension. What do you think?

atolosadelgado: done :)

atolosadelgado:
> Hi Alvaro, excellent! Can you add a test and add a link between sim and digi hits? See e.g. https://github.com/key4hep/k4RecTracker/pull/28/files

I have added the association and the test. I have inspected the output file and the associations seems to make sense. 

Associations are now an additional output collection of the algorithm, but I am not sure if this is the proper way.

As a final remark, using `float` in python or `double` in C++ may introduce systematic differences of 0.2mm, I do not know if it is important. You can use the following script to inspect the output file of the test.

```python
import ROOT
from podio import root_io
import dd4hep as dd4hepModule
from ROOT import dd4hep
from ROOT import TVector3, TFile, TH1D

ofile = TFile("kk.root","update")
hDpw_calc = TH1D("hDpw_calc","",100,0,10)
hDpw_alg  = TH1D("hDpw_alg" ,"",100,0,10)
hDpw_diff = TH1D("hDpw_diff","",100,0,10)

input_file_path = "dch_proton_10GeV_digi.root"
podio_reader = root_io.Reader(input_file_path)

for event in podio_reader.get("events"):
  for dc_hit in event.get("DCH_DigiSimAssociationCollection"):
    digihit =  dc_hit.getDigi()
    simhit  =  dc_hit.getSim()
    digipos = TVector3( digihit.getPosition()[0],digihit.getPosition()[1],digihit.getPosition()[2] )
    simpos  = TVector3(  simhit.getPosition()[0], simhit.getPosition()[1], simhit.getPosition()[2] )
    hit_wire_vector = simpos - digipos
    distance_calculated = hit_wire_vector.Mag()
    distance_from_alg = digihit.getDistanceToWire()
    # print( distance_from_alg - distance_calculated )
    #print( f"{digihit.getEDep()}      {simhit.getEDep()}")
    hDpw_calc.Fill(distance_calculated)
    hDpw_alg.Fill(distance_from_alg)
    hDpw_diff.Fill(distance_calculated-distance_from_alg)
    
ofile.cd()
hDpw_calc.Write()
hDpw_alg.Write()
hDpw_diff.Write()
ofile.Close()
```

atolosadelgado:
@BrieucF  the code is ready for review :)

DCHdigi/src/DCHdigi.cpp
```diff
@@ -0,0 +1,671 @@
...
+}
+
+///////////////////////////////////////////////////////////////////////////////////////
+///////////////////////       finalize       //////////////////////////////////////////
+///////////////////////////////////////////////////////////////////////////////////////
+StatusCode DCHdigi::finalize()
+{
+    if( m_create_debug_histos.value() )
+    {
+        std::unique_ptr<TFile> ofile{TFile::Open ( "dch_digi_alg_debug.root", "recreate" ) };
```
BrieucF, line 197: Can we have the name of the output rootfile for debug th1's configurable from the steering file? (with a default value so that it can also be ignored)

BrieucF (commented):
Thanks Alvaro!

BrieucF:
Let's add the interface as well: https://github.com/BrieucF/k4RecTracker/blob/fix_for_Dolores/DCHdigi/dataFormatExtension/driftChamberHit.yaml#L66

DCHdigi/include/DCHdigi.h
```diff
@@ -0,0 +1,211 @@
...
+ * @param DCH_simhits The name of input collection, type edm4hep::SimTrackerHitCollection <br>
+ * (default name empty) <br>
+ * @param DCH_DigiCollection The name of out collection, type edm4hep::DriftChamberDigiV2Collection <br>
+ * (default name DCH_DigiCollection) <br>
+ * @param DCH_name DCH subdetector name <br>
+ * (default value DCH_v2) <br>
+  * @param fileDataAlg File needed for calculating cluster count and size <br>
+ * (default value /eos/.../DataAlgFORGEANT.root) <br>
+  * @param zResolution_mm Resolution (sigma for gaussian smearing) along the sense wire, in mm <br>
+ * (default value 10 mm) <br>
```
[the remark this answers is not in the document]
atolosadelgado: done

DCHdigi/include/DCHdigi.h
```diff
@@ -0,0 +1,211 @@
...
+  * @param xyResolution_mm Resolution (sigma for gaussian smearing) perpendicular the sense wire, in mm <br>
+ * (default value 10 mm) <br>
```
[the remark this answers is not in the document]
atolosadelgado: done

DCHdigi/include/DCHdigi.h
```diff
@@ -0,0 +1,211 @@
...
+  * @param create_debug_histograms Optional flag to create debug histograms <br>
+ * (default value false) <br>
+  * @param GeoSvcName Geometry service name <br>
+ * (default value GeoSvc) <br>
+  * @param uidSvcName The name of the UniqueIDGenSvc instance <br>
```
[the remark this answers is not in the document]
atolosadelgado: done

DCHdigi/src/DCHdigi.cpp
```diff
@@ -0,0 +1,671 @@
...
+        auto hit_position =  Convert_EDM4hepVector_to_TVector3( input_sim_hit.getPosition(), MM_TO_CM );
+
+        // -------------------------------------------------------------------------
+        //      calculate hit position projection into the wire
+        TVector3 hit_to_wire_vector = this->Calculate_hitpos_to_wire_vector(ilayer, nphi,hit_position);
+        TVector3 hit_projection_on_the_wire = hit_position + hit_to_wire_vector;
+        if( m_create_debug_histos.value() )
+        {
+            double distance_hit_wire = hit_to_wire_vector.Mag();
+            hDpw->Fill(distance_hit_wire);
```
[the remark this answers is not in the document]
BrieucF: Is this going to prevent multi-threading even if the debug option is set to false?

DCHdigi/src/DCHdigi.cpp
```diff
@@ -0,0 +1,671 @@
...
+
+void DCHdigi::PrepareRandomEngine(const edm4hep::EventHeaderCollection&  headers) const
+{
+    uint32_t evt_n = headers[0].getEventNumber();
+    uint32_t run_n = headers[0].getRunNumber();
+    size_t seed = m_uidSvc->getUniqueID(evt_n, run_n, this->name() );
+    m_engine.seed(seed);
+    // test random engine...
+    m_engine.discard(10);
+    myRandom.SetSeed(seed+42);
```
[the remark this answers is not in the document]
atolosadelgado: the mersenne twister may lead to similar results for consecutive seeds (that is, the digitizer might introduce correlations between events). Advancing the internal state some positions should mitigate that issue. I learnt this from @tmadlener, do you know if this is worth it?

The case of ROOT TRandom3 object, they do not provide the method to advance the internal state, so I put the +42 to the seed as a reminder. TRandom3 uses mersenne twister under the hood, so it may be affected by the same problems.

I will crosscheck with Geant4 experts, they had faced the same problem when they moved to MT for sure.

DCHdigi/src/DCHdigi.cpp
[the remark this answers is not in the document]
atolosadelgado: i think everything would run, but the histograms will be partial because they are not summed up when running in MT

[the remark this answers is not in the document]
atolosadelgado: I am not sure why Gaudi functionals do not provide dedicated functions for initialize and terminate workers jobs as other selector structures, see for example [TSelector from ROOT](https://root.cern.ch/doc/master/classTSelector.html).

DCHdigi/src/DCHdigi.cpp
atolosadelgado: done

DCHdigi/include/DCHdigi.h
```diff
@@ -0,0 +1,211 @@
...
+  };
+
+  // the following functions should be upstreamed to the data extension at DD4hep
+  // to avoid code duplication and keep it centralized
+  TVector3 Calculate_hitpos_to_wire_vector(int ilayer, int nphi, const TVector3 & hit_position /*in cm*/) const;
+  TVector3 Calculate_wire_vector_ez       (int ilayer, int nphi) const;
+  TVector3 Calculate_wire_z0_point        (int ilayer, int nphi) const;
+  double   Calculate_wire_phi_z0          (int ilayer, int nphi) const;
+
+  double Calculate_phi_rot_equivalent_to_hit_to_wire_distance(int ilayer, double hit_to_wire_distance) const;
```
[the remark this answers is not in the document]
atolosadelgado: done, removed function because it was unused...

DCHdigi/test/test_DCHdigi/test_DCHdigi.sh
```diff
@@ -0,0 +1,26 @@
...
+fi
+
+# run digitizer for position smearing and cluster counting calculation
+k4run runDCHdigi.py
+
+# check distribution of distance from hit position to the wire
+check_DCHdigi_output=$( (python3 check_DCHdigi_output.py) 2>&1)
+
+# return value printed out by the previous python script
+exit $check_DCHdigi_output
```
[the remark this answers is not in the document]
atolosadelgado: I am sorry, I do not fully understand what it means 'run with source'. After a quick google search, it might be caused by missing shebang, just added it, lets see if that fixes the problem

DCHdigi/include/DCHdigi.h
```diff
@@ -0,0 +1,211 @@
...
+#include "TFile.h"
+#include "TH1D.h"
+
+// Class developed by Walaa for the CLS
+#include "AlgData.h"
+
+/// constant to convert from mm (EDM4hep) to DD4hep (cm)
+constexpr double MM_TO_CM = 0.1;
+
+using colltype_in  = edm4hep::SimTrackerHitCollection;
```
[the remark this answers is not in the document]
atolosadelgado: done, alias removed

DCHdigi/test/test_DCHdigi/test_DCHdigi.sh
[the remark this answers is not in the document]
BrieucF: I mean that when one does `source test_DCHdigi.sh`, the ssh connection closes due to the call to `exit`? If you really need this exit code, run it in a subshell to avoid closing the session

```suggestion
(exit $check_DCHdigi_output)
```

BrieucF:
> > Let's add the interface as well: https://github.com/BrieucF/k4RecTracker/blob/fix_for_Dolores/DCHdigi/dataFormatExtension/driftChamberHit.yaml#L66
> 
> does this do the trick? how can I test that it works? [47b0d5d](https://github.com/key4hep/k4RecTracker/commit/47b0d5db342e1035ff86d85318c1041665c974b9)

I am surprised that it does not complain that the interfaces `edm4hep::TrackerHit:` is then defined twice in edm4hep. Let's ask @tmadlener what he thinks. Context: the goal is to avoid having to redefine `extension::TrackerHit3D` and `extension::TrackerHitPlane` in order to build an interface, as done here https://github.com/BrieucF/k4RecTracker/blob/fix_for_Dolores/DCHdigi/dataFormatExtension/driftChamberHit.yaml#L66. Last time I tried I could not mix `extension` objects and `edm4hep` object in an interface. The latter is indeed very annoying because, in order for subsequent algorithm to work, we have to cast e.g. `edm4hep::TrackerHit3D` coming from the SIM step into `extension::TrackerHit3D` and store this duplicated collection in the output rootfile.

DCHdigi/test/test_DCHdigi/test_DCHdigi.sh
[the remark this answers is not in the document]
tmadlener: If you need the exit code to propagate to the calling shell, you can also simply use the fact, that the return code of the full script will be the return code of the last command. Hence, you can also simply run `python3 check_DCHdigi_output.py > 2>&1` as last command and it will have the same effect.

DCHdigi/src/DCHdigi.cpp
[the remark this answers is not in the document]
atolosadelgado: fixed!

DCHdigi/test/test_DCHdigi/test_DCHdigi.sh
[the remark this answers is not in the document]
atolosadelgado: many thanks, done!

DCHdigi/src/DCHdigi.cpp
```diff
@@ -0,0 +1,664 @@
...
+
+///////////////////////////////////////////////////////////////////////////////////////
+///////////////////////       CalculateNClusters       ////////////////////////////////
+///////////////////////////////////////////////////////////////////////////////////////
+std::pair<uint32_t,uint32_t> DCHdigi::CalculateClusters(const edm4hep::SimTrackerHit & input_sim_hit) const
+{
+    std::pair<uint32_t,uint32_t> return_values = {0,0};
+    uint32_t & Ncl  = return_values.first;
+    uint32_t & ClSz = return_values.second;
+    //_________________SET NECESSARY PARAMETERS FOR THE CLS ALGORITHM-----WALAA_________________//
```
BrieucF, line 344: All these parameters will be set at every call of the function, since many are hard coded, would'nt it be better to define them in the header once and for all?

DCHdigi/src/DCHdigi.cpp
```diff
@@ -0,0 +1,664 @@
...
+   std::vector<float> vecExtraD;
+
+   double Ffrac, Fmpv1, Fsgm1, Fmpv2, Fsgm2;
+   std::vector<double> CorrpMean,CorrpSgm,Corrdgmean,Corrdgsgm,Corrdglfrac,Corrdglmpvl,Corrdglsgml,Corrdglmeang,Corrdglsgmg;
+   float Ekdelta;
+   float maxEx0, maxExSlp, ExSgmlep, ExSgmhad;
+   float MPVEx, SgmEx, MeanEx1, SgmEx1, frac, Slp, CorrSlp, CorrInt;
+   double LengthTrack, Etot_per_track;
+   //////end parameters
+		bool IsSecondaryWithinDCH = false;
```
BrieucF, line 427: Can we fix the indentation?

DCHdigi/dataFormatExtension/driftChamberHit.yaml
```diff
@@ -41,3 +42,101 @@ datatypes:
...
+      - int32_t type                  // type of the raw data hit
+      - int32_t quality               // quality bit flag of the hit
+      - float time                    // time of the hit [ns]
+      - float eDep                    // energy deposited on the hit [GeV]
+      - float eDepError               // error measured on eDep [GeV]
+      - edm4hep::Vector3d position    // point on the sensitive wire (SW) which is closest to the simhit [mm]
+      - edm4hep::Vector3d directionSW // direction of SW
+      - float distanceToWire          // distance hit-wire [mm]
+      - uint32_t clusterCount         // number of clusters associated to this hit
+      - uint32_t clusterSize          // number of electrons per cluster
```
BrieucF, line 60: what about renaming those `nCluster` and `nElectronPerCluster` to be more explicit?

atolosadelgado: done

DCHdigi/src/DCHdigi.cpp
atolosadelgado: done

atolosadelgado:
> > > Let's add the interface as well: https://github.com/BrieucF/k4RecTracker/blob/fix_for_Dolores/DCHdigi/dataFormatExtension/driftChamberHit.yaml#L66
> > 
> > 
> > does this do the trick? how can I test that it works? [47b0d5d](https://github.com/key4hep/k4RecTracker/commit/47b0d5db342e1035ff86d85318c1041665c974b9)
> 
> I am surprised that it does not complain that the interfaces `edm4hep::TrackerHit:` is then defined twice in edm4hep. Let's ask @tmadlener what he thinks. Context: the goal is to avoid having to redefine `extension::TrackerHit3D` and `extension::TrackerHitPlane` in order to build an interface, as done here https://github.com/BrieucF/k4RecTracker/blob/fix_for_Dolores/DCHdigi/dataFormatExtension/driftChamberHit.yaml#L66. Last time I tried I could not mix `extension` objects and `edm4hep` object in an interface. The latter is indeed very annoying because, in order for subsequent algorithm to work, we have to cast e.g. `edm4hep::TrackerHit3D` coming from the SIM step into `extension::TrackerHit3D` and store this duplicated collection in the output rootfile.

I changed the data extension as you proposed originally  (by duplicating code), so the interface can work as expected.

atolosadelgado:
Hi,

I have changed the data extension for this digitizer, so the size of each cluster is given (before the total number of electrons along the step was reported)

Few remarks (maybe worth to put this in readme file?
* It makes a smearing in the position along the wire and radially, according to the input parameters
* Number of clusters and their size is calculated as the option 3 from this paper [1]
* Debug histograms are created if debug option is enabled
* Stand alone test run simulation of the drift chamber based on twisted tubes, and then apply the digitizer. Dedicated directory with all the files needed is given in `DCHdigi/test/test_DCHdigi/`
* It requires that the cellID contain the layer and number of cell within the layer (nphi). It does not matter if the segmentation comes from geometrical segmentation by using twisted tubes and hyperboloids (and the cellID is created out of volume IDs), or the segmentation is virtual DD4hep segmentation.

Some plots:
1. individual cluster size for kaons at 45 GeV
individual_cluster_size_kaon45GeV
2. individual cluster size for protons at 5 GeV
individual_cluster_size_proton5GeV
3. impact of step length, comparison 1mm vs 10 mm (default range cut 0.7mm, for both), protons at 10 GeV with direction (1,1,1), discrepancies on 1-5 electron-size cluster number, otherwise shape is similar.
cluster_size_comparison_protons_10GeV
4. distance of hit to the wire (protons 10GeV, direction 111, max step length 1 mm) 
distance_hit_to_wire

atolosadelgado:
I have added a suffix with the version `v01` to this new digitizer, and writen a readme file in the DCHdigi directory

please let me know if there is anything else I can do :)

DCHdigi/README.md
```diff
@@ -0,0 +1,20 @@
+# Drift chamber (DCH) digitizers
+
+## DCHsimpleDigitizer
```
BrieucF, line 3: this is the description of `DCHsimpleDigitizerExtendedEdm` can you change the "title", add that this version is meant to be used on DriftChamber_o1_v01 from k4geo and add the description of DCHsimpleDigitizer with something like "Algorithm for creating digitized drift chamber hits (based on edm4hep::TrackerHit3D) from edm4hep::SimTrackerHit. You have to specify the expected resolution in z and in xy (distance to the wire). The smearing is applied in the wire reference frame. Deprecated."

DCHdigi/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+* Similar to DCHdigi_v01, but it uses the placement matrix of the wires to apply the smearing along/perpendicular to the sense wire
+* No cluster counting information added into the digitized output
+* It relies on a dedicated data extension, similar to DCHdigi_v01
+
+## DCHdigi description
```
BrieucF, line 9:
```suggestion
## DCHdigi_v01 description
```

DCHdigi/README.md
```diff
@@ -0,0 +1,20 @@
...
+* Number of clusters and their size is derived from file containing precalculated distributions, specified by the parameter `fileDataAlg`, and implemented as the option 3 described in F. Cuna et al, arXiv:2105.07064
+* Debug histograms are created if `create_debug_histograms` option is enabled (output file name can be given)
+* Stand alone test run simulation of the drift chamber based on twisted tubes, and then apply the digitizer. Dedicated directory with all the files needed is given in `DCHdigi/test/test_DCHdigi/`
+* It requires that the cellID contain the layer and number of cell within the layer (nphi). It does not matter if the segmentation comes from geometrical segmentation by using twisted tubes and hyperboloids (and the cellID is created out of volume IDs), or the segmentation is virtual DD4hep segmentation
+* New digitized hit class is used as an EDM4hep data extension, to be integrated into EDM4hep
+* Random number generator uses the seeds calculated on an event basis by the UID service, from the podio header information (run/event number)
+* At the moment, only one version: v01
```
BrieucF, line 20:
```suggestion
```

DCHdigi/README.md
```diff
@@ -0,0 +1,20 @@
+# Drift chamber (DCH) digitizers
+
+## DCHsimpleDigitizer
+
+* Similar to DCHdigi_v01, but it uses the placement matrix of the wires to apply the smearing along/perpendicular to the sense wire
```
BrieucF, line 5: Since `DCHdigi_v01` is only described  later, one should maybe not described this one based on `DCHdigi_v01` (or reverse the order of the README. A description could be from what is written in the .h doxygen description.

DCHdigi/README.md
```diff
@@ -0,0 +1,20 @@
...
+* No cluster counting information added into the digitized output
+* It relies on a dedicated data extension, similar to DCHdigi_v01
+
+## DCHdigi description
+
```
BrieucF, line 10: Can you add that "this digitizer is meant to be used with `DriftChamber_o1_v02` and is expected to work for the upcoming `DriftChamber_o1_v03`"

BrieucF (commented):
Thanks a lot Alvaro! Can you add a plot of "distance to the wire of digitized hit - distance of  the wire from the simhit" together with the code snippet that allows to derive the distance to the wire from the simHit position + cellID (using the facilities from DCH_info)? By the way, can this also be done in Python?

atolosadelgado:
> distance to the wire of digitized hit

Hi,

Thanks for reviewing it!

I am not sure I understood your question:

* `distance to the wire of digitized hit` is zero, by construction: the digitized hit position is a point along the closest wire to the sim hit.
*  `distance of the wire from the simhit`, this information is stored in the output class in the member named `distanceToWire`, [line 58 of the yaml file](https://github.com/key4hep/k4RecTracker/blob/39555a8792f73ba5ad93b0b736a061c72e0cc840/DCHdigi/dataFormatExtension/driftChamberHit.yaml#L58)
* To draw a 2D histogram with the distance to the wire and the cell ID, we can use:
```
events->Draw("DCH_DigiCollection.distanceToWire:DCH_DigiCollection.cellID")
```
For more advanced plots, I guess one would need a short script, which can be in python

BrieucF:
> > distance to the wire of digitized hit
> 
> Hi,
> 
> Thanks for reviewing it!
> 
> I am not sure I understood your question:
> 
>     * `distance to the wire of digitized hit` is zero, by construction: the digitized hit position is a point along the closest wire to the sim hit.

What is "zero" is the distance between the `position` member of the digitizedHit and the wire, by definition. What I am referring to here is given by the member `distanceToWire` of the digitized hit

> 
>     * `distance of the wire from the simhit`, this information is stored in the output class in the member named `distanceToWire`, [line 58 of the yaml file](https://github.com/key4hep/k4RecTracker/blob/39555a8792f73ba5ad93b0b736a061c72e0cc840/DCHdigi/dataFormatExtension/driftChamberHit.yaml#L58)
> 
No, the information you point is for the digitized hit. For the simHit, we need a routine (I thought you had one defined in the DCH_info) that gives the distance to the wire using the CellID and the simHit position.

>     * To draw a 2D histogram with the distance to the wire and the cell ID, we can use:
> 
> 
> ```
> events->Draw("DCH_DigiCollection.distanceToWire:DCH_DigiCollection.cellID")
> ```
> 
> For more advanced plots, I guess one would need a short script, which can be in python

The plot I am asking for is a validation: getDistanceToTheWire(simHit.position, cellID) - digitizedHit.getDistanceToWire. And we should see a gaussian with 100 um. Indirectly this code snippet will also provide an example on how to retrieve this info for simHit.

Is it clearer?

BrieucF:
Attaching here the number of cluster for 10 GeV protons as produced by the tests 
Screenshot 2024-11-21 at 11 26 32 AM

BrieucF:
Thanks Alvaro!

PR #34, opened 2024-09-25
Make sure histograms still work with Gaudi v39

tmadlener:
BEGINRELEASENOTES
- Make sure that histograms keep working / compiling with Gaudi v39

ENDRELEASENOTES


tmadlener:
I have built locally against Gaudi v39r0. Since the nightlies still have v38r3 this should cover everything we need, I think.

jmcarcell:
See https://github.com/key4hep/k4FWCore/pull/239

PR #39, opened 2024-11-25
Create tracks only from charged genParticles in TracksFromGenParticles

Archil-AD:
BEGINRELEASENOTES
- Only a charged genParticle is considered to form a track in TracksFromGenParticles algorithm 

ENDRELEASENOTES


BrieucF:
Thanks Archil!

PR #43, opened 2025-01-15
Changes to run pandora on tracks created from generator-level particles

giovannimarchiori:
Started from Archil's implementation in https://github.com/Archil-AD/k4RecTracker/tree/pandora but
1) changed name of algorithm, make it live in parallel to the transformed based one
2) added the possibility to pass more than one collection of sim tracker hits in input, so that first hit can be calculated from vtx and last hit from wrapper
3) add more debug information
4) do not keep tracks without associated simtrackerhits
5) retrieve B field and detector size automatically
6) implement also extrapolation to calo endcaps

BEGINRELEASENOTES
- Add a Gaudi::Algorithm to create tracks from generator-level particles with charge!=0 and number of simtrackerhits>0 (temporary work around needed because one can not use the edm4hep <--> lcio converters with IOSvc and functionals)
- Track states at IP, first tracker hit, last tracker hit and extrapolation to calorimeter are calculated and saved by the algorithm
ENDRELEASENOTES

Tagging @BrieucF - let's have a look at the code not just in terms of style but also at what it does so that we do things right. For instance might be things that need to be done a bit differently for tracks that point towards the ECAL endcap rather than the barrel. I will make it a draft while we iterate over it


BrieucF:
Hi Giovanni, can you add to the release note that this is a work around needed because currently one can not us the edm4hep <--> lcio converters when using functionals?

giovannimarchiori:
Done

Tracking/components/TracksFromGenParticlesAlg.cpp
```diff
@@ -0,0 +1,241 @@
...
+        for (const auto& hit : *coll) {
+          const edm4hep::MCParticle particle = hit.getParticle();
+          std::array<double,6> ahit{hit.x(), hit.y(), hit.z(), hit.getMomentum()[0], hit.getMomentum()[1], hit.getMomentum()[2]};
+          if(particle.getVertex() == genParticle.getVertex() && particle.getMomentum() == genParticle.getMomentum()) trackHits.push_back(ahit);
+        }
+      }
+    
+      if(!trackHits.empty())
+      {
+	      // particles with at least one SimTrackerHit
```
BrieucF, line 138: is there an indentation problem or just my display?

Tracking/components/TracksFromGenParticlesAlg.cpp
```diff
@@ -0,0 +1,241 @@
...
+	      debug() << "Number of SimTrackerHits: " << trackHits.size() << endmsg;
+	
+        // sort the hits according to radial distance from beam axis
+	      // FIXME: does this work well also for tracks going to endcaps of vtx/wrapper?
```
BrieucF, line 142: I think we have to go in Pandora to see how this should be defined for the endcap... It would likely become the "z" coordinate I guess. If yes, we have to add the GeoSvc and use the flags `DETYPE_ENDCAP`

Tracking/components/TracksFromGenParticlesAlg.cpp
```diff
@@ -0,0 +1,241 @@
...
+        std::sort(trackHits.begin(), trackHits.end(), [](const std::array<double,6> a, const std::array<double,6> b) {
+          double rhoA = std::sqrt(a[0]*a[0] + a[1]*a[1]);
+          double rhoB = std::sqrt(b[0]*b[0] + b[1]*b[1]);
+          return rhoA < rhoB;
+        });
+
+        // TrackState at First Hit
+        auto trackState_AtFirstHit = edm4hep::TrackState {};
+	      auto firstHit = trackHits.front();
```
BrieucF, line 151: Indentation?

Tracking/components/TracksFromGenParticlesAlg.cpp
```diff
@@ -0,0 +1,241 @@
...
+        trackState_AtFirstHit.phi = helixAtFirstHit.getPhi0();
+        trackState_AtFirstHit.omega = helixAtFirstHit.getOmega();
+        trackState_AtFirstHit.Z0 = helixAtFirstHit.getZ0();
+        trackState_AtFirstHit.tanLambda = helixAtFirstHit.getTanLambda();
+        trackState_AtFirstHit.referencePoint = edm4hep::Vector3f((float)posAtFirstHit[0],(float)posAtFirstHit[1],(float)posAtFirstHit[2]);
+        trackFromGen.addToTrackStates(trackState_AtFirstHit);
+
+        // TrackState at Last Hit
+        auto trackState_AtLastHit = edm4hep::TrackState{};
+	      auto lastHit = trackHits.back();
```
BrieucF, line 172: indentation?

Tracking/components/TracksFromGenParticlesAlg.cpp
```diff
@@ -0,0 +1,241 @@
...
+        auto helixAtFirstHit = HelixClass_double();
+        helixAtFirstHit.Initialize_VP(posAtFirstHit, momAtFirstHit, genParticle.getCharge(), m_Bz);
+        // fill the TrackState parameters
+        trackState_AtFirstHit.location = edm4hep::TrackState::AtFirstHit;
+        trackState_AtFirstHit.D0 = helixAtFirstHit.getD0();
+        trackState_AtFirstHit.phi = helixAtFirstHit.getPhi0();
+        trackState_AtFirstHit.omega = helixAtFirstHit.getOmega();
+        trackState_AtFirstHit.Z0 = helixAtFirstHit.getZ0();
+        trackState_AtFirstHit.tanLambda = helixAtFirstHit.getTanLambda();
+        trackState_AtFirstHit.referencePoint = edm4hep::Vector3f((float)posAtFirstHit[0],(float)posAtFirstHit[1],(float)posAtFirstHit[2]);
```
BrieucF, line 167: Does Pandora expect this as the "real" first hit present in the track or is it rather a convention to define also the trackState at the first layer of the tracking system (regardless of whether there is an actual hit or not)? (possibly look at how they are defined for CLD)

BrieucF: The same question holds for the other trackStates

BrieucF (commented):
Thanks Giovanni! Here some comments

Tracking/components/TracksFromGenParticlesAlg.cpp
```diff
@@ -0,0 +1,241 @@
...
+    /// Handle for input MC particles
+    mutable DataHandle<edm4hep::MCParticleCollection> m_inputMCParticles{"MCParticles", Gaudi::DataHandle::Reader, this};
+    /// List of input sim tracker hit collections
+    Gaudi::Property<std::vector<std::string>> m_inputSimTrackerHitCollections{this, "InputSimTrackerHits", {}, "Names of SimTrackerHit collections to read"};
+    /// the vector of input DataHandles for the tracker hit collections
+    std::vector<DataObjectHandleBase*> m_inputSimTrackerHitCollectionHandles;
+    /// Solenoid magnetic fild
+    Gaudi::Property<float> m_Bz{this, "Bz", 2., "Z component of the (assumed constant) magnetic field in Tesla."};
+    /// Inner radius of the ECAL
+    Gaudi::Property<float> m_RadiusAtCalo{this, "RadiusAtCalo", 2172.8, "Inner radius (in mm) of the calorimeter where the TrackState::AtCalorimeter should be defined."};
```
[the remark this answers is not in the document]
giovannimarchiori: ok will try to have a look at this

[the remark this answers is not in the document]
giovannimarchiori: I guess for endcap we would need to know innerZ and also lowerR. But if we retrieve the info from GeoSvc, then we don't need to think about the name of these properties ;)

Tracking/components/TracksFromGenParticlesAlg.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+StatusCode TracksFromGenParticlesAlg::execute(const EventContext&) const {
+    // Get the input MC particle collection
+    const edm4hep::MCParticleCollection* genParticleColl = m_inputMCParticles.get();
+    // Create the output track collection and track gen<->reco links collection
+    auto outputTrackCollection = new edm4hep::TrackCollection();
+    auto MCRecoTrackParticleAssociationCollection = new edm4hep::TrackMCParticleLinkCollection();
+
+    // loop over the gen particles, find charged ones, and create the corresponding reco particles
+    int iparticle = 0;
```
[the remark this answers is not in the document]
giovannimarchiori: iparticle is not to provide a unique identifier, just to print in the debug information the position in the MCParticle vector in case one would like to look at more details in the simulation file.

Tracking/components/TracksFromGenParticlesAlg.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+      // find SimTrackerHits associated to genParticle
+      std::vector<std::array<double,6> > trackHits;
+      for ( size_t ih=0; ih<m_inputSimTrackerHitCollectionHandles.size(); ih++ ) {
+        auto handle = dynamic_cast<DataHandle<edm4hep::SimTrackerHitCollection>*> (m_inputSimTrackerHitCollectionHandles[ih]);
+        const edm4hep::SimTrackerHitCollection* coll = handle->get();
+        for (const auto& hit : *coll) {
+          const edm4hep::MCParticle particle = hit.getParticle();
+          std::array<double,6> ahit{hit.x(), hit.y(), hit.z(), hit.getMomentum()[0], hit.getMomentum()[1], hit.getMomentum()[2]};
+          if(particle.getVertex() == genParticle.getVertex() && particle.getMomentum() == genParticle.getMomentum()) trackHits.push_back(ahit);
```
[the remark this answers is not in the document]
giovannimarchiori: This line was just copied over from Archil's code. I can replace it with your proposed change

Tracking/components/TracksFromGenParticlesAlg.cpp
giovannimarchiori: well, one has to pass to pandora the latest track hit before the ECAL, so that is well defined, I'd say. It's just to think if sorting in radial distance is the most robust way to get the last hit for all tracks

Tracking/components/TracksFromGenParticlesAlg.cpp
giovannimarchiori: to be investigated. Anyway for truth primary tracks and no detector inefficiency the two things are more or less the same (things differ from secondaries of course..)

Tracking/components/TracksFromGenParticlesAlg.cpp
giovannimarchiori: fixed

Tracking/components/TracksFromGenParticlesAlg.cpp
giovannimarchiori: fixed

Tracking/components/TracksFromGenParticlesAlg.cpp
giovannimarchiori: fixed

Tracking/components/TracksFromGenParticlesAlg.cpp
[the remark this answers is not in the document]
giovannimarchiori: Bz implemented (easily retrieved as z of B field at center of detector). Next: radius at calo, I need to think what's better to be used - maybe the info in the layeredcalodata info.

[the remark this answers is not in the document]
giovannimarchiori: done in latest commit

Tracking/components/TracksFromGenParticlesAlg.cpp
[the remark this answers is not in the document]
giovannimarchiori: fixed in latest commit

Tracking/components/TracksFromGenParticlesAlg.cpp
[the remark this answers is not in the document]
giovannimarchiori: implemented in latest commit

Tracking/components/TracksFromGenParticlesAlg.cpp
giovannimarchiori: https://github.com/PandoraPFA/PandoraSDK/blob/f27234ae7dc4a1ebd0445783b4ecdc8b0febe406/include/Objects/Track.h#L272: The track state at the start of the track
So to me this looks like the real first hit.
Being this said, it is not used anywhere in Pandora... https://github.com/search?q=org%3APandoraPFA+m_trackStateAtStart&type=code
neither is the trackStateAtLast.
Discussing at the meeting with Dolores earlier today, it seems for CLD they never set these two track states. I would then keep the current implementation, and revisit later

Tracking/components/TracksFromGenParticlesAlg.cpp
giovannimarchiori: since the truth hits have time information, I used that for sorting them and find first/last hit on the track

Tracking/components/TracksFromGenParticlesAlg.cpp
```diff
@@ -0,0 +1,343 @@
...
+#include "DD4hep/DetType.h"
+#include "DD4hep/DetectorSelector.h"
+
+// C++
+#include <string>
+
+/** @class TracksFromGenParticlesAlg
+ *
+ *  GaudiAlg version of TracksFromGenParticles, that builds an edm4hep::TrackCollection out of an edm4hep::MCParticleCollection.
+ *  It just builds an helix out of the genParticle position, momentum, charge and user defined z component of the (constant) magnetic field.
```
atolosadelgado, line 36: Is is user defined z component of the magnetic field, or it is retrieved from the detector description?

Tracking/CMakeLists.txt
```diff
@@ -4,7 +4,15 @@ project(${PackageName})
 
 #find_package(GenFit)
 find_package(MarlinUtil REQUIRED)
-
+list(APPEND CMAKE_MODULE_PATH $ENV{PANDORAPFA}/cmakemodules)
```
atolosadelgado, line 7: @jmcarcell is it good practice to source cmake find file from other projects or better to copypaste it here?

atolosadelgado (commented):
Thanks for this PR. 

I confirm it compiles. 

@BrieucF do you think it would be worth to provide a test for CI, which can also be used as example for potential users?

Tracking/components/TracksFromGenParticlesAlg.cpp
giovannimarchiori: retrieved - fixed the comment in latest commit

Tracking/CMakeLists.txt
giovannimarchiori: FWIW - I copied this line from k4GaudiPandora https://github.com/key4hep/k4GaudiPandora/blob/4b9f4b92e1e92abc15c594f7c74584f1f2269742/CMakeLists.txt#L31
Looking forward to here from @jmcarcell if that's ok or there's a better solution

BrieucF:
> @BrieucF do you think it would be worth to provide a test for CI, which can also be used as example for potential users?

Yes, I think it is worth adding a test

Tracking/CMakeLists.txt
jmcarcell: Looks good, other package set `CMAKE_MODULE_PATH` too for Pandora packages.

giovannimarchiori:
Test added ;)

giovannimarchiori:
The test fails because for the extrapolation to the calorimeter it requires the calorimeter data extension to be filled, as it will be possible when https://github.com/key4hep/k4geo/pull/419 will be merged. Should we (1) wait for that PR to be merged, (2) merge anyway with a test failing, or (3) add a protection to the algorithm that when the extension is missing the extrapolation to the calo is not performed? (3) seems the best way to me but I will be busy all day...

atolosadelgado:
> The test fails because for the extrapolation to the calorimeter it requires the calorimeter data extension to be filled, as it will be possible when [key4hep/k4geo#419](https://github.com/key4hep/k4geo/pull/419) will be merged. Should we (1) wait for that PR to be merged, (2) merge anyway with a test failing, or (3) add a protection to the algorithm that when the extension is missing the extrapolation to the calo is not performed? (3) seems the best way to me but I will be busy all day...

Hi @giovannimarchiori 

Thanks for adding the test! I just realized that this algorithm relies on the data extension of a specific calorimeter, making it quite specialized and less reusable. Would you consider either renaming it to better reflect its specificity or adapting it to be more general? For example, it could read the size variables [[link](https://github.com/key4hep/k4RecTracker/blob/386bc24d14233848d1ef10fb7739387e84b05333/Tracking/components/TracksFromGenParticlesAlg.cpp#L151)] from input instead of depending on the data extension.

Let me know what you think!

giovannimarchiori:
Hi @atolosadelgado 
the algorithm looks for extension data for calorimeters with standard types
dd4hep::DetType::CALORIMETER | dd4hep::DetType::ELECTROMAGNETIC | dd4hep::DetType::BARREL
dd4hep::DetType::CALORIMETER | dd4hep::DetType::ELECTROMAGNETIC | dd4hep::DetType::ENDCAP
so it is agnostic with respect to the detector model - all detectors should have an EM barrel and endcap. Being this said, I added a protection that if the extensions for these calorimeters are not found, a warning is printed and the trackAtCalo state is not filled. Note that trackAtCalo is needed for p-flow as the impact position at the inner face at the ecal so this is what we really need here - I would not customise the algorithm further. And we prefer to read the positions automatically from the compact file rather than hardcoding them in the scripts to avoid mistakes when copying numbers from one place to another.

atolosadelgado:
> Hi @atolosadelgado the algorithm looks for extension data for calorimeters with standard types dd4hep::DetType::CALORIMETER | dd4hep::DetType::ELECTROMAGNETIC | dd4hep::DetType::BARREL dd4hep::DetType::CALORIMETER | dd4hep::DetType::ELECTROMAGNETIC | dd4hep::DetType::ENDCAP so it is agnostic with respect to the detector model - all detectors should have an EM barrel and endcap. Being this said, I added a protection that if the extensions for these calorimeters are not found, a warning is printed and the trackAtCalo state is not filled. Note that trackAtCalo is needed for p-flow as the impact position at the inner face at the ecal so this is what we really need here - I would not customise the algorithm further. And we prefer to read the positions automatically from the compact file rather than hardcoding them in the scripts to avoid mistakes when copying numbers from one place to another.

Hi @giovannimarchiori 

That makes sense. Since the algorithm depends on the calorimeter, could you reference it in the name? The ARC group will need to develop a similar algorithm without this dependency, so a more explicit name would help avoid confusion. Thanks!

atolosadelgado:
Currently, the data extension is only used to extract the size of the calorimeter. Would it be possible to add an option (as a non-default setting) to pass the calorimeter size as parameters of the Gaudi algorithm? This would make the algorithm more flexible and suitable for other systems that do not provide the data extension.

PR #45, opened 2025-02-13
Retrieve B field and detector info in TracksFromGenParticlesWithECalExtrap functional

giovannimarchiori:
add retrieval of B field and extrapolation to calorimeter to functional that produces tracks from gen particles.
Based on transformer code from TracksFromGenParticles.cpp and on detector retrieval and extrapolation code from [racksFromGenParticlesWithECalExtrapAlg.cpp.

BEGINRELEASENOTES
- add retrieval of B field and extrapolation to calorimeter to functional that produces tracks from gen particles
ENDRELEASENOTES

### DIFF
--- a/Tracking/components/TracksFromGenParticlesWithECalExtrap.cpp
+++ b/Tracking/components/TracksFromGenParticlesWithECalExtrap.cpp
@@ -1,0 +1,326 @@
+#include "Gaudi/Property.h"
+
+// edm4hep
+#include "edm4hep/MCParticleCollection.h"
+#include "edm4hep/SimTrackerHitCollection.h"
+#include "edm4hep/TrackCollection.h"
+#include "edm4hep/TrackMCParticleLinkCollection.h"
+
+// marlin
+#include <marlinutil/HelixClass_double.h>
+
+// pandora
+#include <Objects/Helix.h>
+
+// k4FWCore
+#include "k4FWCore/Transformer.h"
+
+// DD4HEP
+#include <DDRec/DetectorData.h>
+#include "DD4hep/Detector.h"
+#include "DD4hep/DD4hepUnits.h"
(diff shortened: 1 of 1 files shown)


giovannimarchiori:
rebased

atolosadelgado:
hi @giovannimarchiori 

Thanks for this PR! I’m curious, am I correct in understanding that the main difference from the algorithm introduced three weeks ago is that the previous one was a Gaudi algorithm, while this PR introduces the same functionality as a Gaudi functional?

If that’s the case, do you think it would be beneficial to wrap the core logic of the Gaudi algorithm into a dedicated class? That way, we could avoid code duplication.

Thanks again for your work!

giovannimarchiori:
Hi @atolosadelgado,
before doing that - is there a reason on the long term to keep both versions? If at some point the Gaudi::Algorithm should just be phased out, then I am not sure it's worth spending time refactoring the code.
I don't know the answer myself, just let me know

atolosadelgado:
> Hi @atolosadelgado, before doing that - is there a reason on the long term to keep both versions? If at some point the Gaudi::Algorithm should just be phased out, then I am not sure it's worth spending time refactoring the code. I don't know the answer myself, just let me know

As in the TV show 'who wants to be a millionaire', I will play the joker card 'phone a friend': @BrieucF @kjvbrt @jmcarcell , do you think it is worth to refactor code, or it is ok to duplicate it as Gaudi algorithm is going to be deprecated in the near future?

giovannimarchiori:
Hi @atolosadelgado 
I created an issue https://github.com/key4hep/k4RecTracker/issues/46 to keep in mind "to remove the Gaudi::Algorithm once we are 'migrated' to functionals and Gaudi::Algorithm is phased out"
Thanks, Giovanni

PR #49, opened 2025-02-27
Fix rng seeding and drawing random numbers

m-fila:
BEGINRELEASENOTES
- Use `getUniqueID` directly with `EventHeaderCollection` to avoid truncating casts
- Add seeding of TRandom3 rng in DCHdigi_v01. Make TF1 draw random numbers from instad of global rng.

ENDRELEASENOTES

This is an update to use `getUniqueID` directly with `EventHeaderCollections` - added in https://github.com/key4hep/k4FWCore/pull/287. The code as is had a truncating cast for event number. To avoid possible mistakes, `EventHeaderCollections` can be used directly so there is no need to extract manually event number and run number

I also spoted that `myRandom` was never seeded. The TF1 were also drawing numbers from global rng rather than one of the local rngs. Let me know if that was done on purpose

### DIFF
--- a/DCHdigi/src/DCHdigi_v01.cpp
+++ b/DCHdigi/src/DCHdigi_v01.cpp
@@ -285,8 +285,10 @@ void DCHdigi_v01::PrintConfiguration(std::ostream& io) {
 }
 
 void DCHdigi_v01::PrepareRandomEngine(const edm4hep::EventHeaderCollection& headers) const {
-  size_t   seed  = m_uidSvc->getUniqueID(headers, this->name());
-  m_engine.seed(seed);
+  auto engine_seed = m_uidSvc->getUniqueID(headers, this->name());
+  m_engine.seed(engine_seed);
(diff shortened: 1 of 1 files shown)


atolosadelgado:
Hi @m-fila 

thanks for the update and fixes! 

I see that the CI is failing with this error,
```
 CMake Error at CMakeLists.txt:21 (find_package):
  Could not find a configuration file for package "podio" that is compatible
  with requested version "1.2".
```

shall we wait until it is fixed?

atolosadelgado (approved):

BrieucF:
Hi @m-fila , thanks a lot indeed! Can you point to the documentation on how to properly use the random number generator?

m-fila:
Currently we have only this [page](https://key4hep.github.io/key4hep-doc/developing-key4hep-software/k4fwcore/doc/uniqueIDGen.html) But there is a PR to update it  https://github.com/key4hep/k4FWCore/pull/288

Issue #48, opened 2025-02-27
Drift Chamber digitizer returns 0 for distanceFromTheWireError and positionAlongTheWireError


andread3vita:
### Check duplicate issues.

- [ ] Checked for duplicates

### Goal

Hi,
I am using the digitizer defined in k4RecTracker/DCHdigi and if I read the results I get that distanceFromTheWireError and positionAlongTheWireError return 0.

Here is what I am doing:

```
import ROOT
from podio import root_io
import dd4hep as dd4hepModule
from ROOT import dd4hep

input_file_path = "output_digi_v3.root"
podio_reader = root_io.Reader(input_file_path)
for i, event in enumerate(podio_reader.get("events")):
    
    digi_dc = event.get("DCH_DigiCollection")
    
    for digi_hit in digi_dc:
        
        digi_hit_DistanceToWireError = digi_hit.getDistanceToWireError()
        digi_hit_PositionAlongWireError = digi_hit.getPositionAlongWireError()
        
        print(digi_hit_DistanceToWireError,digi_hit_PositionAlongWireError)
```

Thank you!
Andrea

### Operating System and Version

Alma Linux 9.5

### compiler

GCC 14.2.0

### The version of the key4hep stack

2025-02-12

### Package Version

current master

### Reproducer

[output_digi_v3.root.zip](https://github.com/user-attachments/files/19009196/output_digi_v3.root.zip)

atolosadelgado:
Hi @andread3vita 

thank you for using the issue tracking system.  I will open a PR to fix it.

atolosadelgado:
Hi @andread3vita 

Could you please check if https://github.com/key4hep/k4RecTracker/pull/50 fixes the bug?

Thank you for your time!

andread3vita:
Hi @atolosadelgado 
Yes! This fixes the bug.

Thank you!

PR #50, opened 2025-02-27
DCH digi now fills DistanceToWireError and PositionAlongWireError fields 

atolosadelgado:
BEGINRELEASENOTES
- DCH digi output now contains correct values for DistanceToWireError and PositionAlongWireError. Before it was zero. Fixes https://github.com/key4hep/k4RecTracker/issues/48

ENDRELEASENOTES


atolosadelgado:
@jmcarcell CI for release stack seems broken for a different reason. The change introduced here is small, could I still proceed and merge this change?

PR #51, opened 2025-02-28
Set test dependencies for tracking tests that need a simulated file

jmcarcell:
BEGINRELEASENOTES
- Set test dependencies for tracking tests that need a simulated file

ENDRELEASENOTES

This should help with the failures in CI, that currently may or may not fail depending if these two tests run before the one that generates their input file or not.


atolosadelgado:
out of curiosity, is this dependency honored when running the tests in parallel, for instance `ctest -j8`?

PR #52, opened 2025-03-04
Move rdn distributions inside call operator

atolosadelgado:
BEGINRELEASENOTES
- Limit scope of random distributions inside the call operator to prevent race conditions

ENDRELEASENOTES

The following random distributions have an internal state, and therefore making them just mutable is not enough to prevent a race condition. Thanks @m-fila for noticing!

```
Standard random distributions in `<random>` that **store an internal state**:

1. **`std::normal_distribution`**  
   - Uses the **Box-Muller transform** (or Ziggurat method, implementation-dependent) to generate normally distributed numbers.
   - Box-Muller requires generating two uniform random numbers to compute two normal variates at once. Since only one may be returned per call, the second is stored for the next invocation.

2. **`std::lognormal_distribution`**  
   - Based on `std::normal_distribution`, so it inherits the same internal state.

3. **`std::chi_squared_distribution`**  
   - Often implemented using a sum of squared normal-distributed numbers, which may use an internal state.

4. **`std::cauchy_distribution`**  
   - Sometimes uses an internal state to handle extreme values in its heavy-tailed distribution.

5. **`std::gamma_distribution`**  
   - Uses an internal state for more complex number generation algorithms like the Marsaglia-Tsang method.

6. **`std::weibull_distribution`**  
   - May cache intermediate calculations for efficiency.

7. **`std::extreme_value_distribution`**  
   - Sometimes uses an internal state for transformations.

The key reason these distributions maintain state is that they involve **transforms or algorithms that produce multiple values at a time** but can only return one per function call. The remaining value(s) need to be stored for subsequent calls to maintain efficiency.
```

### DIFF
--- a/DCHdigi/src/DCHdigi_v01.cpp
+++ b/DCHdigi/src/DCHdigi_v01.cpp
@@ -31,11 +31,9 @@ StatusCode DCHdigi_v01::initialize() {
 
   if( 0 > m_z_resolution.value() )
     ThrowException("Z resolution input value can not be negative!");
-  m_gauss_z_cm  = std::normal_distribution<double>(0., m_z_resolution.value() * MM_TO_CM);
 
   if( 0 > m_xy_resolution.value() )
     ThrowException("Radial (XY) resolution input value can not be negative!");
-  m_gauss_xy_cm = std::normal_distribution<double>(0., m_xy_resolution.value() * MM_TO_CM);
 
   //-----------------
   // Retrieve the subdetector
@@ -101,6 +99,10 @@ DCHdigi_v01::operator()(const edm4hep::SimTrackerHitCollection& input_sim_hits,
                     const edm4hep::EventHeaderCollection&   headers) const {
   // initialize seed for random engine
   this->PrepareRandomEngine(headers);
+  // Gaussian random number generator used for the smearing of the z position, in cm!
+  std::normal_distribution<double> m_gauss_z_cm{0., m_z_resolution.value() * MM_TO_CM};
+  // Gaussian random number generator used for the smearing of the xy position, in cm!
+  std::normal_distribution<double> m_gauss_xy_cm{0., m_xy_resolution.value() * MM_TO_CM};
 
   debug() << "Input Sim Hit collection size: " << input_sim_hits.size() << endmsg;
 
--- a/DCHdigi/include/DCHdigi_v01.h
+++ b/DCHdigi/include/DCHdigi_v01.h
@@ -127,12 +127,6 @@ private:
   inline static thread_local std::mt19937_64 m_engine;
   void                                       PrepareRandomEngine(const edm4hep::EventHeaderCollection& headers) const;
 
-  // Operator std::normal_distribution<T>::operator()(Generator& g) is a non-const member function and thus cannot be called for a constant object. So we defined the distribution as mutable.
-  // Gaussian random number generator used for the smearing of the z position, in cm!
-  mutable std::normal_distribution<double> m_gauss_z_cm;
-  // Gaussian random number generator used for the smearing of the xy position, in cm!
-  mutable std::normal_distribution<double> m_gauss_xy_cm;
-
   /// members with internal state (such as random engines) must be defined thread local
   inline static thread_local TRandom3 myRandom;
   //------------------------------------------------------------------


DCHdigi/src/DCHdigi_v01.cpp
```diff
@@ -101,6 +99,10 @@ DCHdigi_v01::operator()(const edm4hep::SimTrackerHitCollection& input_sim_hits,
                     const edm4hep::EventHeaderCollection&   headers) const {
   // initialize seed for random engine
   this->PrepareRandomEngine(headers);
+  // Gaussian random number generator used for the smearing of the z position, in cm!
+  std::normal_distribution<double> m_gauss_z_cm{0., m_z_resolution.value() * MM_TO_CM};
+  // Gaussian random number generator used for the smearing of the xy position, in cm!
+  std::normal_distribution<double> m_gauss_xy_cm{0., m_xy_resolution.value() * MM_TO_CM};
```
m-fila, line 105:
```suggestion
  std::normal_distribution<double> gauss_z_cm{0., m_z_resolution.value() * MM_TO_CM};
  // Gaussian random number generator used for the smearing of the xy position, in cm!
  std::normal_distribution<double> gauss_xy_cm{0., m_xy_resolution.value() * MM_TO_CM};
```

You might want to rename them since "m_"  prefix you probably reserve for the member variables

m-fila (commented):
I have a single suggestion on naming. Otherwise looks good

atolosadelgado:
@m-fila 

thanks for reviewing!

In your [newer documentation](https://github.com/key4hep/k4FWCore/pull/288/files#diff-2bd079d855ea3412490ccdfd78c8636d9ca61ebb56390a4db8725b5535f611e1), you suggest to also recreate the random number engine inside the call operator. Is it the recommended approach then? I can modify that as well

m-fila (approved):
> In your [newer documentation](https://github.com/key4hep/k4FWCore/pull/288/files#diff-2bd079d855ea3412490ccdfd78c8636d9ca61ebb56390a4db8725b5535f611e1), you suggest to also recreate the random number engine inside the call operator. Is it the recommended approach then? I can modify that as well

I think it's fine the way it is. The example in the docs is very simple, for a bigger class using RNG in many different parts of the code I'd consider making the RNG member too.

atolosadelgado:
> > In your [newer documentation](https://github.com/key4hep/k4FWCore/pull/288/files#diff-2bd079d855ea3412490ccdfd78c8636d9ca61ebb56390a4db8725b5535f611e1), you suggest to also recreate the random number engine inside the call operator. Is it the recommended approach then? I can modify that as well
> 
> I think it's fine the way it is. The example in the docs is very simple, for a bigger class using RNG in many different parts of the code I'd consider making the RNG a member too.

Thanks! Another question, currently the random engine is marked as `inline static`. But if the same algorithm is called twice in the workflow, it can lead to a race condition right?

atolosadelgado:
> > Thanks! Another question, currently the random engine is marked as `inline static`. But if the same algorithm is called twice in the workflow, it can lead to a race condition right?
> 
> We talked about this in-person. I was mistaken about `thread_local` implying `static` for members. It looks that `thread_local` members **must** be also `static`, `thread_local` implies `static` for variables declared in a block scope (inside function).
> 
> `static thread_local` members normaly would be problematic when multiple instances of the class are be created because they will be shared among all the instances. But in this particular case the generator is seeded for each algorithm execution, so there is no need to care about sharing an information between multiple instances. If the instances were to execute in parallel, then they'd do so in different thread and do to `thread_local` each thread has its own generator, so there shouldn't be any problems with multi-threading
> 
> ~It'd be better to get rid of `thread_local` members too as you originally suggested~ I still think it's fine but you may remove it, just to avoid all these considerations. Sorry for the confusion

Many thanks! 

@BrieucF is it ok for you?

BrieucF:
Is there no way to declare the random generator in `initialize` AND be thread safe?

atolosadelgado:
> Is there no way to declare the random generator in `initialize` AND be thread safe?

A random engine (which generates random numbers between 0 and 1), should be `thread_local` member of the class. A priori, one does not know the parallelization scheme, so makes no sense to mark it as static. A static random engine ca lead to race conditions if 2 instances are created and run in parallel.

Regarding the distributions, since they have an internal state, one has to mark them as  `thread_local` member of the class. Alternatively, one can limit their scope, as @m-fila suggested in his example, as it is implemented in this PR.

From my side, it is ready to go :)

m-fila:
> I am checking if removing the static actually works...

`thread_local` members have to be also `static`

> Is there no way to declare the random generator in `initialize` AND be thread safe?

In principle this could be possible but there are other downsides. The most trivial would be protecting the rng with a mutex, but then getting a random number becomes time-expensive operation. Another possibility would be to have a separate service doing something smart like pre-generating a set of numbers


> A random engine (which generates random numbers between 0 and 1), should be `thread_local` member of the class. A priori, one does not know the parallelization scheme, so makes no sense to mark it as static. A static random engine ca lead to race conditions if 2 instances are created and run in parallel.

1. Distributions have a state so slapping `mutable` on top of them is not thread-safe. Making them `thread_local` and initializing in `initialize` won't work as that would initialize only single thread. Making them a local to `execute` looks like a good idea.
2. Generator can be `static thread_local` members since the "initialization" with seed is happening during algorithm execution. There is no data-race here as each thread has it's own rng and a thread runs a single algorithm at a time.  Or they could be made local variable to the execution.

DCHdigi/src/DCHdigi_v01.cpp
```diff
@@ -99,8 +97,23 @@ StatusCode DCHdigi_v01::initialize() {
 std::tuple<extension::SenseWireHitCollection, extension::SenseWireHitSimTrackerHitLinkCollection>
 DCHdigi_v01::operator()(const edm4hep::SimTrackerHitCollection& input_sim_hits,
                     const edm4hep::EventHeaderCollection&   headers) const {
+  /// initialize engines
+  std::mt19937_64 m_engine;
```
m-fila, line 101:
```suggestion
  std::mt19937_64 rng_engine;
```
Prefix `m_` is probably reserved for members

m-fila (commented):
Sorry to be this picky, I have another suggestion regarding naming. 
You migh also keep the `PrepareRandomEngine` if you like to have this part in a separate function. Just keep it's signature and tweak a little
```cpp
std::tuple<std::mt19937_64, TRandom3> DCHdigi_v01::PrepareRandomEngine(const edm4hep::EventHeaderCollection& headers) const {
  auto engine_seed = m_uidSvc->getUniqueID(headers, this->name());
  auto rng_engine = std::mt19937_64(engine_seed);
  auto random_seed = m_uidSvc->getUniqueID(headers, this->name()+"_1");
  auto myRandom = TRandom3(random_seed);
  // advance internal state to minimize possibility of creating correlations
  rng_engine.discard(10);
  for (int i = 0; i < 10; ++i)
    myRandom.Rndm();
  return {rng_engine, myRandom};
}
```

DCHdigi/src/DCHdigi_v01.cpp
atolosadelgado: thanks! changed!

atolosadelgado:
> Sorry to be this picky, I have another suggestion regarding naming. You migh also keep the `PrepareRandomEngine` if you like to have this part in a separate function. Just keep it's signature and tweak a little
> 
> ```c++
> std::tuple<std::mt19937_64, TRandom3> DCHdigi_v01::PrepareRandomEngine(const edm4hep::EventHeaderCollection& headers) const {
>   auto engine_seed = m_uidSvc->getUniqueID(headers, this->name());
>   auto rng_engine = std::mt19937_64(engine_seed);
>   auto random_seed = m_uidSvc->getUniqueID(headers, this->name()+"_1");
>   auto myRandom = TRandom3(random_seed);
>   // advance internal state to minimize possibility of creating correlations
>   rng_engine.discard(10);
>   for (int i = 0; i < 10; ++i)
>     myRandom.Rndm();
>   return {rng_engine, myRandom};
> }
> ```

thanks! I have changed as you suggested!

m-fila (approved):
LGTM

jmcarcell:
I rebased and removed a few formatting changes since these should be done in a single commit so that we can add it to `.git-blame-ignore-revs`. We should also add pre-commit here. After this is merged I think there is nothing urgent and I can tag this repository.

PR #57, opened 2025-03-28
clean up debug output

giovannimarchiori:
BEGINRELEASENOTES
- Print out full particle details only in verbose mode, add succinct summary in debug mode

ENDRELEASENOTES

### DIFF
--- a/Tracking/components/TracksFromGenParticles.cpp
+++ b/Tracking/components/TracksFromGenParticles.cpp
@@ -181,17 +181,31 @@ struct TracksFromGenParticles final
 
     // loop over the gen particles, find charged ones, and create the corresponding reco particles
     int iparticle = 0;
-    for (const auto& genParticle : genParticleColl) {
+    for (const edm4hep::MCParticle& genParticle : genParticleColl) {
       edm4hep::Vector3d p = genParticle.getMomentum();
       double pmag = std::sqrt(p.x*p.x + p.y*p.y + p.z*p.z);
       debug() << endmsg;
-      debug() << "Gen. particle: " << genParticle << endmsg;
-      debug() <<"  particle "<<iparticle++<<"  PDG: "<< genParticle.getPDG()  << " momentum: "<< pmag
-              << " charge: "<< genParticle.getCharge() << endmsg;
-      debug() << "Particle decayed in tracker: " << genParticle.isDecayedInTracker() << endmsg;
+      verbose() << "Gen. particle: " << genParticle << endmsg;
+      debug() << "Particle " << iparticle++
+              << "  identifier: " << genParticle.getObjectID()
+              << "  PDG: " << genParticle.getPDG()
+              << "  momentum: " << pmag
+              << "  charge: " << genParticle.getCharge() << endmsg;
+      debug() << "  generator status: " << genParticle.getGeneratorStatus()
+              << "  simulator status: " << genParticle.getSimulatorStatus() << endmsg;
+      {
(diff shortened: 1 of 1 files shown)


Tracking/components/TracksFromGenParticles.cpp
```diff
@@ -181,17 +181,31 @@ struct TracksFromGenParticles final
...
+      verbose() << "Gen. particle: " << genParticle << endmsg;
+      debug() << "Particle " << iparticle++
+              << "  identifier: " << genParticle.getObjectID()
+              << "  PDG: " << genParticle.getPDG()
+              << "  momentum: " << pmag
+              << "  charge: " << genParticle.getCharge() << endmsg;
+      debug() << "  generator status: " << genParticle.getGeneratorStatus()
+              << "  simulator status: " << genParticle.getSimulatorStatus() << endmsg;
+      {
+        auto daughters = genParticle.getDaughters();
```
BrieucF, line 197: Shall we encapsulate this in an if statement like `if (msgLevel()>= MSG::DEBUG)`?

Tracking/components/TracksFromGenParticles.cpp
```diff
@@ -181,17 +181,31 @@ struct TracksFromGenParticles final
...
       double pmag = std::sqrt(p.x*p.x + p.y*p.y + p.z*p.z);
       debug() << endmsg;
-      debug() << "Gen. particle: " << genParticle << endmsg;
-      debug() <<"  particle "<<iparticle++<<"  PDG: "<< genParticle.getPDG()  << " momentum: "<< pmag
-              << " charge: "<< genParticle.getCharge() << endmsg;
-      debug() << "Particle decayed in tracker: " << genParticle.isDecayedInTracker() << endmsg;
+      verbose() << "Gen. particle: " << genParticle << endmsg;
+      debug() << "Particle " << iparticle++
+              << "  identifier: " << genParticle.getObjectID()
+              << "  PDG: " << genParticle.getPDG()
```
BrieucF, line 191: Isn't this (and others) already printed by `<< genParticle` from the `verbose` line above? (verbose will also be printed in debug mode)

giovannimarchiori: no, it's the opposite, verbose is more verbose than debug - that's why I want to move the genParticle to verbose to avoid printing out that very detailed information in "normal" debug output

https://github.com/key4hep/Gaudi/blob/54cdeb940467934ad1c63b3c943342b4d8907d4a/GaudiKernel/python/GaudiKernel/Constants.py#L28

```
# message levels -----------------------------------------------------------
ALL = 0
VERBOSE = 1
DEBUG = 2
INFO = 3
WARNING = 4
ERROR = 5
FATAL = 6
```

Tracking/components/TracksFromGenParticles.cpp
giovannimarchiori: it should not be necessary given the way the message stream work (if level is smaller than debug these messages won't appear in output even without the check)

BrieucF: no but, right now, aren't you still doing some unnecessary manipulation e.g. when the output level is set to `INFO`?

Tracking/components/TracksFromGenParticles.cpp
BrieucF: Indeed, got confuse by other conventions, sorry!

PR #58, opened 2025-04-15
FIX um -> mm

andread3vita:
BEGINRELEASENOTES
- Corrected the unit in the comment describing spatial resolution (`VTXdigi/include/VTXdigitizer.h`): previously stated as µm, but the correct unit is mm.

ENDRELEASENOTES


atolosadelgado:
Thanks for your PR @andread3vita !

@armin-ilg could you please crosscheck?

PR #62, opened 2025-06-16
Case handling to enable the track extrapolation for IDEA

SanghyunKo:
IDEA o1 is using a unified detector description for the whole barrel and endcap, hence added the case handling to enable the track extrapolation.

BEGINRELEASENOTES
- Case handling to enable the track (from GenPtc) extrapolation for IDEA

ENDRELEASENOTES

### DIFF
--- a/Tracking/components/TracksFromGenParticles.cpp
+++ b/Tracking/components/TracksFromGenParticles.cpp
@@ -78,9 +78,12 @@ struct TracksFromGenParticles final
     if (m_extrapolateToECal) {
       bool retrieved = false;
 
-      try {
-        const dd4hep::rec::LayeredCalorimeterData * caloExtension = getExtension( ( dd4hep::DetType::CALORIMETER | dd4hep::DetType::BARREL | dd4hep::DetType::ENDCAP),
-                                                                                  ( dd4hep::DetType::AUXILIARY | dd4hep::DetType::FORWARD ) );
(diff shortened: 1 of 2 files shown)


Tracking/components/TracksFromGenParticles.cpp
```diff
@@ -101,35 +101,62 @@ struct TracksFromGenParticles final
...
+      bool retrieved = false;
+
       try {
-        const dd4hep::rec::LayeredCalorimeterData * eCalBarrelExtension = getExtension( ( dd4hep::DetType::CALORIMETER | dd4hep::DetType::ELECTROMAGNETIC | dd4hep::DetType::BARREL),
-                                                                                        ( dd4hep::DetType::AUXILIARY  |  dd4hep::DetType::FORWARD ) );
-        m_eCalBarrelInnerR = eCalBarrelExtension->extent[0] / dd4hep::mm;
-        m_eCalBarrelMaxZ = eCalBarrelExtension->extent[3] / dd4hep::mm;
-        debug() << "ECAL barrel extent: Rmin [mm] = " << m_eCalBarrelInnerR << endmsg;
-        debug() << "ECAL barrel extent: Zmax [mm] = " << m_eCalBarrelMaxZ << endmsg;
+        const dd4hep::rec::LayeredCalorimeterData * caloExtension = getExtension( ( dd4hep::DetType::CALORIMETER | dd4hep::DetType::BARREL | dd4hep::DetType::ENDCAP),
```
jmcarcell, line 107: Does this always throw when it's not found? If not we could check the returned pointer which should be `nullptr`? The several try - except look a bit ugly

PR #76, opened 2026-02-24
remove compatibility with older Gaudi versions

m-fila:
BEGINRELEASENOTES
- Remove histogram interface compatibility with Gaudi versions older that v39

ENDRELEASENOTES

Both the nigthlies and the current release (2026-02-01) have Gaudi  newer than 39.

See https://github.com/key4hep/k4FWCore/pull/378


tmadlener (approved):
One CI workflow is failing for unrelated reasons.

PR #78, opened 2026-04-07
IDEA DCH full digitization chain

muhammadsahil5030:
BEGINRELEASENOTES

## Description

This PR introduces a full digitization chain for the IDEA Drift Chamber (DCH).

The implementation follows the standard detector signal formation sequence:
ionization → drift → amplification → signal formation → electronics → digitization.

* New Gaudi algorithm:

  * `DCHdigi_v03`

    * cluster generation based on Bethe–Bloch
    * drift time from 2D LUT (x–t relation)
    * Polya-distributed avalanche gain
    * waveform formation (single-electron pulse model)
    * signal propagation along wire (delay + attenuation)
    * electronics response (transfer function)
    * ADC digitization

* Drift time handling:

  * `DCHXT2DLUT`

    * reads `xt_mean` and `xt_error` from ROOT file
    * provides interpolated drift time + smearing

* Noise simulation:

  * `DCHFFTNoise`

    * FFT-based colored noise generation
    * compatible with waveform sampling

* EDM4hep extensions:

  * `SenseWireHit`
  * waveform storage via `RawTimeSeriesCollection`

* Steering:

  * configurable via `runDCHdigiV3.py`

## Outputs

* `SenseWireHitCollection`
* `SenseWireHitSimTrackerHitLinkCollection`
* Digitized waveforms (left/right) as `RawTimeSeriesCollection`

ENDRELEASENOTES


DCHdigi/test/test_DCHdigi/runDCHdigiV3.py
```diff
@@ -0,0 +1,93 @@
+#
```
jmcarcell, line 1: There should be a test that runs this steering file, otherwise we don't know if it runs.

DCHdigi/include/BetheBloch.h
```diff
@@ -0,0 +1,65 @@
...
+
+//double getDelta(double p, double m){
+//	double delta;
+//	double x= TMath::Log10(p/m);
+//
+//	cout<<x<<"\t"<<p<<endl;
+//	if(x>=x1) delta=k2*x-c;
+//	if(x<x1&&x>=x0) delta=k2*x-c+a*pow((x1-x),k1);
+//	if(x<x0) delta=0.0;
+//
```
tmadlener, line 41: Are these used anywhere else than below in `bethe_bloch`? In that case I would simply move them there. Also these could / should probably marked `constepr` (where possible) or at least `const`

DCHdigi/include/BetheBloch.h
```diff
@@ -0,0 +1,65 @@
...
+//	return delta;
+//}
+
+double bethe_bloch( double *xp,double *par){
```
tmadlener, line 45: It doesn't look like this has to go through a ROOT interface at any point. Hence, we should be able to pass the `xp` and `par` as `,std::array`s. Checking the only call site of this, it might even just be `double xp` and `std::array<double, 2> par`. Alternatively already explicitly use `m` and `I` as arguments to not hide behind a way too generic interface.

DCHdigi/include/DCHdigi_v03.h
```diff
@@ -0,0 +1,476 @@
...
+	      const edm4hep::EventHeaderCollection&)> {
+
+  DCHdigi_v03(const std::string& name, ISvcLocator* svcLoc);
+
+  StatusCode initialize() override;
+  StatusCode finalize() override;
+
+  std::tuple<
+	  extension::SenseWireHitCollection, 
+	  extension::SenseWireHitSimTrackerHitLinkCollection,
```
tmadlener, line 152: Should this still use the extension types, or rather the ones that went to EDM4hep?

DCHdigi/include/DCHdigi_v03.h
```diff
@@ -0,0 +1,476 @@
...
+#include <memory>
+
+
+//requirments for the xtrelTIME
+//class XTRELTIME; //uncommnent when needed
+
+class TF1;
+class TFile;
+
+/// constant to convert from mm (EDM4hep) to DD4hep (cm)
```
tmadlener, line 131: Stray comment?

DCHdigi/include/DCHdigi_v03.h
```diff
@@ -0,0 +1,476 @@
...
+  }
+
+  int CalculateNphiFromCellID(dd4hep::DDSegmentation::CellID id) const { return m_decoder->get(id, "nphi"); }
+
+  TVector3 Convert_EDM4hepVector_to_TVector3(const edm4hep::Vector3d& v, double scale) const {
+    return {v[0] * scale, v[1] * scale, v[2] * scale};
+  };
+  edm4hep::Vector3d Convert_TVector3_to_EDM4hepVector(const TVector3& v, double scale) const {
+    return {v.x() * scale, v.y() * scale, v.z() * scale};
+  };
```
tmadlener, line 226: At least for the scaling we have utilities in EDM4hep: https://github.com/key4hep/EDM4hep/blob/6d24c90b72c2434cacf3824a9b05b758d7e3911f/utils/include/edm4hep/utils/vector_utils.h#L293-L296

I.e. it should be possible to do:
```cpp
return edm4hep::Vector3d{v.x(), v.y(), v.z()} * scale;
```

DCHdigi/include/DCHdigi_v03.h
```diff
@@ -0,0 +1,476 @@
...
+  /// Detector name
+  Gaudi::Property<std::string> m_DCH_name{this, "DCH_name", "DCH_v2", "Name of the Drift Chamber detector"};
+
+  /// Pointer to the geometry service
+  SmartIF<IGeoSvc> m_geoSvc;
+
+  mutable Gaudi::Accumulators::Counter<Gaudi::Accumulators::atomicity::full, unsigned int> m_event_counter;
+
+  /// Decoder for the cellID
+  dd4hep::DDSegmentation::BitFieldCoder* m_decoder;
```
tmadlener, line 180: This should probably be a smart pointer.

DCHdigi/include/DCHdigi_v03.h
```diff
@@ -0,0 +1,476 @@
...
+      "Spatial resolution in the z direction (from reading out the wires at both sides) in mm. Default 1 mm."};
+  /// xy resolution in mm
+  Gaudi::Property<float> m_xy_resolution{this, "xyResolution_mm", 0.1,
+                                         "Spatial resolution in the xy direction in mm. Default 0.1 mm."};
+
+  /// create seed using the uid
+  SmartIF<IUniqueIDGenSvc> m_uidSvc;
+
+  /// Create random engine, initialized with seed out of Event Header
+  std::tuple<std::mt19937_64, TRandom3> CreateRandomEngines(const edm4hep::EventHeaderCollection& headers) const;
```
tmadlener, line 202: What is the reason here to have two random engines at work?

DCHdigi/src/DCHdigi_v03.cpp
```diff
@@ -0,0 +1,1381 @@
...
+
+std::tuple<extension::SenseWireHitCollection,
+	extension::SenseWireHitSimTrackerHitLinkCollection,
+	
+	//use raw time series collection:
+	edm4hep::RawTimeSeriesCollection,
+	edm4hep::RawTimeSeriesCollection>
+
+DCHdigi_v03::operator()(const edm4hep::SimTrackerHitCollection& input_sim_hits,
+                    const edm4hep::EventHeaderCollection&   headers) const
```
tmadlener, line 396: This is currently almost 900 lines long(!). Is there a chance to split it up into a few smaller functions with reduced scope to make it slightly easier to digest?

DCHdigi/include/DCHdigi_v03.h
```diff
@@ -0,0 +1,476 @@
...
+	  "Overall scale factor for FFT noise (multiplies the generated time-noise)"};
+  
+  // Remove DC (baseline) component from the generated noise waveform.
+  Gaudi::Property<bool> m_noiseRemoveDC{this, "NoiseRemoveDC", true,
+	  "Remove DC offset from generated noise (recommended)"};
+  // Cached FFT-noise template (loaded once in initialize)
+  mutable bool m_fftNoiseReady{false};
+  mutable std::vector<double> m_fftMag;
+  mutable double m_fftAmpNorm{1.0};     // fft_amp (normalization from file)
+  mutable double m_fftMaxFreq{0.0};     // max frequency from fft_freq (same units as 1/ns)
```
tmadlener, line 392: Do these really have to be `mutable`? The comment implies that they are only mutated in `initialize`. If they need to be `mutable` are the points where they are mutated appropriately guarded?

DCHdigi/include/DCHUtilities.h
```diff
@@ -0,0 +1,218 @@
...
+  double xp[1]  = { p_GeV };                           // GeV (header multiplies by 1e3 internally)
+  double par[2] = { m_GeV * 1.0e3,                     // mass in MeV
+                    m_MeanExcEnergy_eV.value() * 1e-6  // I in MeV
+                  };
+  const double dEdx_MeVcm2_per_g = BB::bethe_bloch(xp, par);
+  const double dEdx_MeV_per_cm = dEdx_MeVcm2_per_g * m_GasDensity_g_cm3.value();
+  const double lambda_per_cm = (dEdx_MeV_per_cm * 1.0e6) / m_W_eff_eV.value();
+
+  return (lambda_per_cm > 0.0) ? lambda_per_cm : 0.0;
+}
```
tmadlener, line 42: This should probably live in the `cpp` file of the algorithm to which it belongs.

DCHdigi/include/DCHUtilities.h
```diff
@@ -0,0 +1,218 @@
...
+    if (s_cm >= l_cm) break;     // stop if we exceed the step length
+    pos_mm.push_back(10.0 * s_cm); // store position in mm
+  }
+  return pos_mm;
+
+}
+
+// --- Experimental cluster size probabilities w(n) for He-iSobutane (90:10) ---
+// Based on Fischle et al., NIM A301 (1991)
+inline const std::vector<double>& w_cluster_He_iC4H10()
```
tmadlener, line 79: This can be done as `constexpr` when using a `std::array<double, 35>`. That way we don't have to keep track of `init` and we don't need a `static vector`. Plus we can guarantee that all the (admittedly probably negligible) work is done at compile time.

See e.g. https://godbolt.org/z/GsPTsG39E

DCHdigi/include/DCHXT2DLUT.h
```diff
@@ -0,0 +1,40 @@
...
+public:
+  DCHXT2DLUT() = default;
+  ~DCHXT2DLUT();
+
+  // Load xt_mean and xt_error from ROOT file (par.root)
+  bool load(const std::string& rootFile,
+            const std::string& meanName  = "xt_mean",
+            const std::string& sigmaName = "xt_error");
+
+  bool isLoaded() const { return m_loaded; }
```
tmadlener, line 22: The way this is used you could get rid of `load` and in turn `isLoaded`. In the algorithm you construct it and immediately afterwards call `load`. You never call `load` again. So you could just make the constructor do the loading in which case you can guarantee that either you have a valid object or you throw an exception in case you don't.

`m_file` should probably be a `unique_ptr` in which case you could save some manual management.

DCHdigi/src/DCHdigi_v03.cpp
```diff
@@ -0,0 +1,1381 @@
...
+std::tuple<std::mt19937_64, TRandom3> DCHdigi_v03::CreateRandomEngines(
+                const edm4hep::EventHeaderCollection& headers) const {
+  auto engine_seed = m_uidSvc->getUniqueID(headers, this->name());
+  auto rng_engine = std::mt19937_64(engine_seed);
+  auto random_seed = m_uidSvc->getUniqueID(headers, this->name()+"_1");
+  auto myRandom = TRandom3(random_seed);
+  // advance internal state to minimize possibility of creating correlations
+  rng_engine.discard(10);
+  for (int i = 0; i < 10; ++i)
+    myRandom.Rndm();
```
tmadlener, line 84: You will probably have to do more than `10` iterations here if you really want to guarantee no correlations. I don't know exactly how many, but I seem to remember O(1000).

DCHdigi/src/DCHdigi_v03.cpp
```diff
@@ -0,0 +1,1381 @@
...
+StatusCode DCHdigi_v03::initialize() 
+{
+  if (!m_uidSvc)
+    ThrowException("Unable to get UniqueIDGenSvc");
+
+  if (0 > m_z_resolution.value())
+    ThrowException("Z resolution input value can not be negative!");
+
+  if (0 > m_xy_resolution.value())
+    ThrowException("Radial (XY) resolution input value can not be negative!");
```
tmadlener, line 100: Many of these can probably be done with `K4_GAUDI_CHECK`. Which would also make it possible to get rid of `ThrowException` (for which I don't see the added value, since `throw std::runtime_error("....")`; is only a small number of additional characters, but withotu hiding what you actually do.

tmadlener (commented):
This doesn't look like it has been `clang-format`ted yet (and CI will likely fail also due to that.

I have started to go through the code and there are a few things that could be improved. I didn't look at all of it, but some of the comments might apply on a broader scope.

DCHdigi/include/DCHdigi_v03.h
s6anloes, line 1: In this file I will be tagging places where I think the code needs to be more generic, since we will want to use the digitiser for the straw tubes as well.
The final renaming of the file/algorithm will need to wait until we have merged Stefano's PR on the `WireTracker_info` class:
https://github.com/key4hep/k4geo/pull/581

DCHdigi/include/DCHdigi_v03.h
```diff
@@ -0,0 +1,476 @@
...
+ * @date   2026-02
+ *
+ * <h4>Input collections and prerequisites</h4>
+ * Processor requires a collection of SimTrackerHits <br>
+ * This code uses DD4hep length natural unit (cm), but EDM4hep data is (usually) in mm. Please be careful with units. <br>
+ *
+ * <br> <h4>Output</h4>
+ * Processor produces digitized hits and optional waveform collections. <br>
+ *
+ * @param DCH_simhits The name of input collection, type edm4hep::SimTrackerHitCollection <br>
```
s6anloes, line 14:
```suggestion
 * @param InputSimHitCollection The name of input collection, type edm4hep::SimTrackerHitCollection <br>
```

DCHdigi/include/DCHdigi_v03.h
```diff
@@ -0,0 +1,476 @@
...
+ * (default name empty) <br>
+ *
+ * @param DCH_DigiCollection The name of output collection, type extension::SenseWireHitCollection <br>
```
s6anloes, line 17:
```suggestion
 * @param OutputDigiHitCollection The name of output collection, type extension::SenseWireHitCollection <br>
```

DCHdigi/include/DCHdigi_v03.h
```diff
@@ -0,0 +1,476 @@
...
+ * (default name DCH_DigiCollection) <br>
+ *
+ * @param DCH_DigiSimAssociationCollection The name of output association collection, type extension::SenseWireHitSimTrackerHitLinkCollection <br>
```
s6anloes, line 20:
```suggestion
 * @param DigiSimAssociationCollection The name of output association collection, type extension::SenseWireHitSimTrackerHitLinkCollection <br>
```

DCHdigi/include/DCHdigi_v03.h
```diff
@@ -0,0 +1,476 @@
...
+ * This code uses DD4hep length natural unit (cm), but EDM4hep data is (usually) in mm. Please be careful with units. <br>
+ *
+ * <br> <h4>Output</h4>
+ * Processor produces digitized hits and optional waveform collections. <br>
+ *
+ * @param DCH_simhits The name of input collection, type edm4hep::SimTrackerHitCollection <br>
+ * (default name empty) <br>
+ *
+ * @param DCH_DigiCollection The name of output collection, type extension::SenseWireHitCollection <br>
+ * (default name DCH_DigiCollection) <br>
```
s6anloes, line 18:
```suggestion
 * (default name SenseWire_DigiCollection) <br>
```

DCHdigi/include/DCHdigi_v03.h
```diff
@@ -0,0 +1,476 @@
...
+ *
+ * @param DCH_DigiSimAssociationCollection The name of output association collection, type extension::SenseWireHitSimTrackerHitLinkCollection <br>
+ * (default name DCH_DigiSimAssociationCollection) <br>
```
s6anloes, line 21:
```suggestion
 * (default name SenseWire_DigiSimAssociationCollection) <br>
```

DCHdigi/include/DCHdigi_v03.h
```diff
@@ -0,0 +1,476 @@
...
+ *
+ * @param DCH_name DCH subdetector name <br>
+ * (default value DCH_v2) <br>
```
s6anloes, line 24:
```suggestion
 * @param DetectorName subdetector name <br>
 * (default value DCH_v2) <br>
```

DCHdigi/include/DCHdigi_v03.h
```diff
@@ -0,0 +1,476 @@
...
+#include "k4Interface/IUniqueIDGenSvc.h"
+
+// EDM4HEP
+#include "edm4hep/EventHeaderCollection.h"
+#include "edm4hep/SimTrackerHitCollection.h"
+#include "edm4hep/RawTimeSeriesCollection.h"
+
+// EDM4HEP extension
+#include "extension/SenseWireHitCollection.h"
+#include "extension/SenseWireHitSimTrackerHitLinkCollection.h"
```
s6anloes, line 93:
```suggestion
#include "SenseWireHitCollection.h"
#include "TrackerHitSimTrackerHitLinkCollection.h"
```
You should not be using the extension definition of SenseWireHit but rather the central one

DCHdigi/include/DCHdigi_v03.h
```diff
@@ -0,0 +1,476 @@
...
+// DD4hep
+#include "DD4hep/Detector.h" // for dd4hep::VolumeManager
+#include "DDSegmentation/BitFieldCoder.h"
+
+// STL
+#include <random>
+#include <string>
+
+// data extension for detector DCH_v2
+#include "DDRec/DCH_info.h"
```
s6anloes, line 108: To be replaced once Stefano's PR is merged

DCHdigi/include/DCHdigi_v03.h
```diff
@@ -0,0 +1,476 @@
...
+#include "TRandom3.h"
+#include "TVector3.h"
+#include "TH1F.h"
+#include "TH2F.h"
+#include "TH3F.h"
+#include "TTree.h"
+#include "TProfile.h"
+#include "TGraph.h"
+#include "DCHXT2DLUT.h"
+#include <memory>
```
s6anloes, line 122: These should be placed in the right category

DCHdigi/include/DCHdigi_v03.h
```diff
@@ -0,0 +1,476 @@
...
+
+
+//requirments for the xtrelTIME
+//class XTRELTIME; //uncommnent when needed
```
s6anloes, line 126: This should not be toggled via commenting and re-compiling.
Can we introduce a new Gaudi::Property flag `use_XTRELTIME` instead?

DCHdigi/include/DCHdigi_v03.h
```diff
@@ -0,0 +1,476 @@
...
+
+class TF1;
+class TFile;
+
+/// constant to convert from mm (EDM4hep) to DD4hep (cm)
+
+struct DCHdigi_v03 final
+    : k4FWCore::MultiTransformer<
+          std::tuple<extension::SenseWireHitCollection, 
+	  extension::SenseWireHitSimTrackerHitLinkCollection,
```
s6anloes, line 136:
```suggestion
	  TrackerHitSimTrackerHitLinkCollection,
```

DCHdigi/include/DCHdigi_v03.h
s6anloes: Indeed, everywhere it should be changed to the central EDM4hep SenseWireHit Collections

DCHdigi/include/DCHdigi_v03.h
```diff
@@ -0,0 +1,476 @@
...
+
+  //------------------------------------------------------------------
+  //          machinery for geometry
+
+  /// Geometry service name
+  Gaudi::Property<std::string> m_geoSvcName{this, "GeoSvcName", "GeoSvc", "The name of the GeoSvc instance"};
+  Gaudi::Property<std::string> m_uidSvcName{this, "uidSvcName", "uidSvc", "The name of the UniqueIDGenSvc instance"};
+
+  /// Detector name
+  Gaudi::Property<std::string> m_DCH_name{this, "DCH_name", "DCH_v2", "Name of the Drift Chamber detector"};
```
s6anloes, line 172: Also here change all the DCH specific names to something more generic

DCHdigi/include/DCHdigi_v03.h
```diff
@@ -0,0 +1,476 @@
...
+  Gaudi::Property<double> m_signalSpeed_mm_per_ns{this, "signalSpeed_mm_per_ns", 200.0,
+	  "Propagation speed along sense wire [mm/ns]. 200 mm/ns ~ 5 ns/m"};
+
+  //Enable or disable the attenuation
+  Gaudi::Property<bool> m_enableWireAttenuation{this, "EnableWireAttenuation", true,
+          "Enable signal attenuation along the sense wire (charge-division effect)"};
+
+  //Attenuation in signal
+  Gaudi::Property<double> m_wireAttenuationLength_mm{this, "WireAttenuationLength_mm", 1.0e4,
+	  "Attenuation length lambda [mm] for exp(-d/lambda). Very large disables attenuation"};
```
s6anloes, line 323: Instead of having two Gaudi::Properties you could also make it such that negative values for the attenuation length toggle it off (e.g. setting the value to -1.0). This removes one redundant parameter

DCHdigi/include/DCHdigi_v03.h
```diff
@@ -0,0 +1,476 @@
...
+  /// Pointer to drift chamber data extension
+  dd4hep::rec::DCH_info* dch_data = {nullptr};
+
+  double m_halfChamberLength_mm = 0.0;	// gas half-length along z
+
+  //------------------------------------------------------------------
+  //          machinery for smearing the position
+
+  /// along the sense wire position resolution in mm
+  Gaudi::Property<float> m_z_resolution{
```
s6anloes, line 191:
```suggestion
  Gaudi::Property<float> m_z_resolution_mm{
```
Since you have units in the other properties I suggest to have them here as well

DCHdigi/include/DCHdigi_v03.h
```diff
@@ -0,0 +1,476 @@
...
+      this, "zResolution_mm", 1.0,
+      "Spatial resolution in the z direction (from reading out the wires at both sides) in mm. Default 1 mm."};
+  /// xy resolution in mm
+  Gaudi::Property<float> m_xy_resolution{this, "xyResolution_mm", 0.1,
```
s6anloes, line 195:
```suggestion
  Gaudi::Property<float> m_xy_resolution_mm{this, "xyResolution_mm", 0.1,
```

DCHdigi/src/DCHdigi_v03.cpp
```diff
@@ -0,0 +1,1381 @@
...
+	edm4hep::RawTimeSeriesCollection,
+	edm4hep::RawTimeSeriesCollection>
+
+DCHdigi_v03::operator()(const edm4hep::SimTrackerHitCollection& input_sim_hits,
+                    const edm4hep::EventHeaderCollection&   headers) const 
+{
+
+  if (m_event_counter.value()%1 == 0)
+  {
+	  info() << "Processing Events: " << m_event_counter.value() <<endmsg;
```
s6anloes, line 401: Seems a bit overkill to have this printout every event? You could change it to every 100 events or so. Alternatively change it to `debug()`.

DCHdigi/src/DCHdigi_v03.cpp
```diff
@@ -0,0 +1,1381 @@
...
+    }
+
+//=======================================================================================
+// 			Accumulate electrons per cellID				       //
+//======================================================================================= 			
+    // Create per-cell accumulator for this cellID
+    auto& cellAccum = cellAccumMap[static_cast<uint64_t>(cellid)];
+
+    // Accumulate total deposited energy summed over all contributing sim hits in the cell
+    cellAccum.eDep += static_cast<float>(input_sim_hit.getEDep());
```
s6anloes, line 777:
```suggestion
```
This line should not be there, since later in the if-else statement there is the accumulation already

DCHdigi/src/DCHdigi_v03.cpp
```diff
@@ -0,0 +1,1381 @@
...
+	  std::vector<double> wfRight(nBinsAnalogWf, 0.0);
+	  std::vector<double> wfElecL(nBinsAnalogWf, 0.0);
+	  std::vector<double> wfElecR(nBinsAnalogWf, 0.0);
+
+
+	  const float WireStereoAngle = acc.wireStereoAngle;
+	  // Find earliest electron arrival time in cell
+	  auto it_min = std::min_element(acc.times_ns.begin(), acc.times_ns.end());
+	  // Find earliest electron arrival time in cell
+	  const double t_hit_ns = (*it_min < 0.0) ? 0.0 : *it_min;
```
s6anloes, line 852:
```suggestion
	  const double t_hit_ns = *it_min;
```
The digitiser should accept negative times as well, since there is the Overlay algorithm that can assign negative time to simhits

DCHdigi/src/DCHdigi_v03.cpp
```diff
@@ -0,0 +1,1381 @@
...
+  int ilayer;
+  int nphi;
+
+  // loop over hit collection
+  int loop_index = 0;
+
+  // Store only one waveform for debuging
+  m_waveformFilled = false;
+  // Local analog waveform definition used for ALL cells
+  const double wfTmin = 0.0;
```
s6anloes, line 435: I have commented below, but it's also relevant here:
The digitiser should also be able to handle time windows smaller than 0, because there is an algorithm that does an overlay of simhits, and it will assign negative times to some of the hits.
The start time here could be a new `Gaudi::Property`

s6anloes (commented):
Hi Muhammad,
thanks for this excellent work! We have discussed some of comments in our last meeting, and I don't have that many more.
Some general things are: i) try to keep "DCH" out of the variable names, we will want to use the digitiser also for the straw tubes, and for this we will also have to rename the algorithm; ii) in some places you protect againt negative timestamps for the digitised hits, but I think this should be removed to work with the overlay algorithm

One conceptual thing on the length and strucutre of the code:
Conceptually, your digitiser does two things:
1) DigiHit creation: One part of the code does all the necessary calculation to create and fill the DigiHit (`SenseWireHit`) directly from the information of the SimHits. In particular I'm talking about the calculation of the hit position and the cluster sizes. This information goes straight into the output collection of `SenseWireHits`
2) Waveform creation: the second part of the code creates the full waveform with noise and electronic response

After the cell accumulation loop, these two parts are fairly independent: the waveform is not calculated from the DigiHits, nor are the digihits directly derived from the waveform.
These two parts are conceptually very different things:
1) the DigiHit creation is essentially an update of the current `DCHdigi_v02` with better cluster calculation, but ultimately it is still a simplified digitiser where the DigiHit information is calculated directly from the SimHits.
2) The waveform creation is a much more significant step towards realistic digitisation (in the sense that this is what a real detector would see). So this waveform creation must be followed by algorithms which take this waveform as input and analyse it to get to the DigiHit with things like the cluster position and size. I know that in INFN these algorithms have been worked on a lot, so at the moment only the porting to Gaudi is missing.

Both of these things are very good and important, but I suspect that in the review a discussion will arise on how we can split this algorithm into two Functionals because it is quite long at the moment. I believe this could be a natural point: split everything that has to do with the waveform creation and everything that has to do with the digihit calculation. The only slight problem with that is that the loop over the simhits to accumlate the hits will be duplicated, but this is probably a small price.

I hope I have made myself clear. Maybe there is no need for this discussion, but these were just my thoughts on this.

In any case, thanks again, this looks really good already.

DCHdigi/src/DCHdigi_v03.cpp
giovannimarchiori: It seems you use the event counter only to printout the number of event being processed, is it? so a natural question is why do you need a dedicated event counter at all as member of your class? You can just use the centrally provided one in your digi script with 

```
from Configurables import EventCounter
eventCounter = EventCounter("EventCounter",
                            OutputLevel=INFO,
                            Frequency=10)
TopAlg += [eventCounter]
```

or whatever.

DCHdigi/include/BetheBloch.h
```diff
@@ -0,0 +1,65 @@
+#pragma once
```
jmcarcell, line 1: There should be some comment about where this implementation is coming from (PDG?).

PR #82, opened 2026-05-08
Remove the deprecated k4DataSvc and PodioOutput

jmcarcell:
BEGINRELEASENOTES
- Remove the deprecated k4DataSvc and PodioOutput, following https://github.com/key4hep/k4FWCore/pull/392

ENDRELEASENOTES


jmcarcell:
Changes look good to me, CI is green. The writer is not run with auditors anymore, but that's probably the least interesting part, or if anyone wants it I can have a look.

PR #84, opened 2026-06-04
Add MUONdigitizer: cluster-based digitizer for IDEA muon system

mahmoudali2:
BEGINRELEASENOTES

- Creating a cluster-based digitization algorithm for IDEA Muon-system
- It clusters the simHits located in the same uRWELL chamber within specific cluster-size (number of strips fired by avalanche) into one digiHit, the value we get from test-beam results.
- The goal is to cluster the secondaries (mainly delta-rays) lies within specific radius away from primary into one cluster, and not many separate digiHits.
- The code also doing the same task as DDPlanarDigi for smearing hits positions and time, and applying efficiency filter.
- It was better to have all digi steps of muon-system in one, instead of calling many different algorithms in steering files because of the requirements of specific types of inputs and associations, and to avoid complexity. 
- There's aslo a forcing hits into surface to avoid hit positions outside the chamber surfaces.
- We expect further developments in that MUONDigi for extra digitization  steps in future

ENDRELEASENOTES

### DIFF
--- a/MUONdigi/include/MUONDigitizer.h
+++ b/MUONdigi/include/MUONDigitizer.h
@@ -19,17 +19,17 @@
 
 // DD4HEP
 #include "DD4hep/Detector.h"
-#include "DDRec/Vector3D.h"
 #include "DDRec/SurfaceManager.h"
+#include "DDRec/Vector3D.h"
 #include "DDSegmentation/BitFieldCoder.h"
 
-#include <vector>
 #include <algorithm>
 #include <cmath>
 #include <cstdint>
 #include <limits>
 #include <map>
 #include <set>
+#include <vector>
 
 /** @class MUONDigitizer
  *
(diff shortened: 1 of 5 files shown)


MUONdigi/src/MUONDigitizer.cpp
```diff
@@ -0,0 +1,353 @@
...
+
+      const double smearedLocalPos_cm[3] = {0., simLocalPos_cm[1] + m_gauss_u.shoot() * dd4hep::mm,
+                                            simLocalPos_cm[2] + m_gauss_v.shoot() * dd4hep::mm};
+
+      double digiGlobalPos_cm[3] = {0., 0., 0.};
+      xform.LocalToMaster(smearedLocalPos_cm, digiGlobalPos_cm);
+
+      // ---- Optional: project smeared digi onto the actual sensitive surface
+      // when it falls outside the chamber's surface bounds.
+      if (m_forceHitsOntoSurface && m_surfMap) {
```
jmcarcell:
```suggestion
      if (m_forceHitsOntoSurface) {
```
`m_surfMap` is always valid, you check for it before, otherwise this line wouldn't be reached

jmcarcell:
Can we find a better name than `MUONdigitizer`? We discussed this in a call recently and our current names are not great. We agreed that a good name would give some indication of what the algorithm actually does. I would say forget about `DDPlanarDigi` and let's think about a completely new name. `MUONdigitizer` sounds like it's the default for muons for all detectors.

mahmoudali2:
> Can we find a better name than `MUONdigitizer`? We discussed this in a call recently and our current names are not great. We agreed that a good name would give some indication of what the algorithm actually does. I would say forget about `DDPlanarDigi` and let's think about a completely new name. `MUONdigitizer` sounds like it's the default for muons for all detectors.

Hi @jmcarcell , thanks for your comments.  
For the name what about `MuonSystemDigitizer` ?
I'd like to hear also from @BrieucF and @s6anloes

MUONdigi/include/MUONDigitizer.h
```diff
@@ -0,0 +1,122 @@
...
+ *       distinct Z-strip indices both within MaxClusterSize. This matches
+ *       the test-beam cluster-size definition (strips per view).
+ *    3. Emit one digi hit at the energy-weighted centroid plus optional
+ *       Gaussian smearing, apply per-cluster efficiency, and repeat.
+ *
+ *  @author Mahmoud Althakeel
+ *  @date   2026-06
+ */
+
+class MUONDigitizer : public Gaudi::Algorithm {
```
tmadlener, line 56: Is there a reason this is not done via a functional algorithm?

ueinhaus:
For the name, how about uRWellDigitiser, or something similar that specifies the technology used?

tmadlener:
Very much in favor of a name that carries the technology along.

PR #85, opened 2026-06-30
Tracking: Fix track reconstruction issues; Jacobian units, hit-track associations and TGeo navigator reset

mahmoudali2:
BEGINRELEASENOTES

- Corrected unit definitions in the Jacobian to ensure consistency with GenFit and EDM4hep.
- Fixed broken associations between hits and tracks after the filtering stage.
- Refactored the track finding algorithm to properly discard hits labeled as noise; only successfully reconstructed tracks are now stored in the output file.
- Reset the global TGeoManager navigator before fitting
- Use hit position as reference point for trackState 2 and 3

ENDRELEASENOTES

### DIFF
--- a/Tracking/components/GenfitTrackFitter.cpp
+++ b/Tracking/components/GenfitTrackFitter.cpp
@@ -315,6 +315,11 @@ struct GenfitTrackFitter final
     // Loop over the tracks created by the pattern recognition step
     for (const auto& track : tracks_input) {
 
+      // Guard the whole per-track fit: a GenFit exception (e.g. an
+      // ill-conditioned covariance during the fit or the extrapolation to the
+      // IP) should skip just this track, not abort the entire event loop.
+      try {
+
       num_tracks += 1;
 
       // Skip unmatched tracks if the option is enabled
@@ -404,6 +409,13 @@ struct GenfitTrackFitter final
                          m_runCalorimeterExtrapolation.value());
           }
         }
+      }
+
+      } catch (const std::exception& e) {
+        number_failures += 1;
+        warning() << "Track " << num_tracks - 1
+                  << ": GenFit exception during fit/extrapolation, skipping track: " << e.what() << endmsg;
+        continue;
       }
     }
 
--- a/Tracking/components/GenfitTrackFitter.cpp
+++ b/Tracking/components/GenfitTrackFitter.cpp
@@ -320,96 +320,97 @@ struct GenfitTrackFitter final
       // IP) should skip just this track, not abort the entire event loop.
       try {
 
-      num_tracks += 1;
+        num_tracks += 1;
 
-      // Skip unmatched tracks if the option is enabled
-      // Consider unmatched tracks those with type = 0
-      if (m_ListOfTypesToSkip.size() > 0 && std::find(m_ListOfTypesToSkip.begin(), m_ListOfTypesToSkip.end(),
-                                                      track.getType()) != m_ListOfTypesToSkip.end()) {
-        num_skip += 1;
-        warning() << "Skipping track " << num_tracks - 1 << " with type " << track.getType() << "\n" << endmsg;
-        continue; // skip unmatched tracks
-      }
-
-      if (track.getTrackerHits().size() < 3) {
-        num_skip += 1;
-        warning() << "Track " << num_tracks - 1 << ": less than 3 hits, skipping fit.\n" << endmsg;
-        continue; // skip tracks with less then 3 hits (seed initialization needs 3 hits)
-      }
-
-      num_processed_tracks += 1;
-
-      int isSuccess = 0;
-      if (m_singleEvaluation) {
-
-        isSuccess = ProcessTrack(track, false, m_particleHypothesis[0], FittedTracks, FittedTracksWithFilteredHits,
-                                 FittedHits, m_runCalorimeterExtrapolation.value());
-
-      } else {
-
-        int winning_hypothesis = FindBestHypothesis(track, false);
-
-        if (winning_hypothesis == -1) {
-          debug() << "Track " << num_tracks - 1 << ": fit failed for all hypotheses, trying with less hits." << endmsg;
-        } else {
-
-          isSuccess = 1;
-          int pdgCode = winning_hypothesis;
-          ProcessTrack(track, false, pdgCode, FittedTracks, FittedTracksWithFilteredHits, FittedHits,
-                       m_runCalorimeterExtrapolation.value());
+        // Skip unmatched tracks if the option is enabled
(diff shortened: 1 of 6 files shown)


Tracking/components/GenfitTrackFitter.cpp
```diff
@@ -315,95 +315,108 @@ struct GenfitTrackFitter final
...
-      if (m_ListOfTypesToSkip.size() > 0 && std::find(m_ListOfTypesToSkip.begin(), m_ListOfTypesToSkip.end(),
-                                                      track.getType()) != m_ListOfTypesToSkip.end()) {
-        num_skip += 1;
-        warning() << "Skipping track " << num_tracks - 1 << " with type " << track.getType() << "\n" << endmsg;
-        continue; // skip unmatched tracks
-      }
+      // Guard the whole per-track fit: a GenFit exception (e.g. an
+      // ill-conditioned covariance during the fit or the extrapolation to the
+      // IP) should skip just this track, not abort the entire event loop.
+      try {
```
jmcarcell: Can your try-except only cover the line that can throw an exception? You probably don't want to catch an unrelated exception, but if it happens in the big block inside the try-except you may.

mahmoudali2: Hi @jmcarcell, it's not only specific one line, but it happens when it calls `ProcessTrack`, which it happens multiple times through the loop. 
But we can change the exception to `genfit::Exception` instead of `std::exception`, so it caches narrower and only related exceptions. What do you think?

jmcarcell: Then it should probably be handled inside `ProcessTrack` and not outside of it, as any caller will be in the same situation. Different return codes can be returned, for example, to distinguish different cases. `using genfit::Exception` is definitely better.

Tagging @andread3vita

andread3vita:
Thank you @mahmoudali2! Do you have a reproducer for the bug you are describing?

andread3vita:
> Thank you @mahmoudali2! Do you have a reproducer for the bug you are describing?

While testing the fitter, I checked whether the failure of a single track fit could cause the entire event processing to crash. To prevent this, I wrapped the fitting procedure in a try-catch block so that a failed fit for one track should not interrupt the processing of the rest of the event. See [here](https://github.com/key4hep/k4RecTracker/blob/8973e465cb6bfc762765b3d324e3ebc8656107d8/Tracking/src/genfit_interfaces/GenfitTrack.cpp#L916).

mahmoudali2:
> > Thank you @mahmoudali2! Do you have a reproducer for the bug you are describing?
> 
> While testing the fitter, I checked whether the failure of a single track fit could cause the entire event processing to crash. To prevent this, I wrapped the fitting procedure in a try-catch block so that a failed fit for one track should not interrupt the processing of the rest of the event. See [here](https://github.com/key4hep/k4RecTracker/blob/8973e465cb6bfc762765b3d324e3ebc8656107d8/Tracking/src/genfit_interfaces/GenfitTrack.cpp#L916).

Hi @andread3vita .. your try/catch guards the fit itself (processTrackWithRep), but the crash happens after the fit succeeds, when the fitted state is retrieved. getFittedState() .. 

To reproduce the error, I've run the fitting `k4run Tracking/test/testTrackFitter/runTestTrackFitter.py`  on the output of the finder from  z-> 2 muons decay here's a direct link to the input sim file : https://cernbox.cern.ch/files/spaces/eos/user/m/maali/IDEA_muon_optimization/vertexing_test/reco/z2mumu_sim.edm4hep.root. 
, and this is the error : 
```
EventLoopMgr         FATAL .executeEvent(): Standard std::exception thrown by k4FWCore__Sequencer
EventLoopMgr         ERROR genfit::Exception thrown with excString:
KalmanFitterInfo::calcAverageState: ill-conditioned covariance matrix.
in line: 149 in file: .../core/src/MeasuredStateOnPlane.cc
.. 
ApplicationMgr       ERROR Application Manager Terminated with error code 6
```

mahmoudali2:
@jmcarcell ..Now I've removed it from the loop and just guard exactly the calls at the source that throws the exception and with `genfit::Exception` ..

andread3vita:
Hi! With the latest commits, the issue should now be resolved without relying on a try-catch block. @mahmoudali2, could you please test if it works for you?

This PR also includes a few other commits with fixes I spotted last week. Additionally, I've made a change to the track finder so that it no longer outputs unclustered tracks.

jmcarcell:
> Hi! With the latest commits, the issue should now be resolved without relying on a try-catch block. @mahmoudali2, could you please test if it works for you?
> 
> This PR also includes a few other commits with fixes I spotted last week. Additionally, I've made a change to the track finder so that it no longer outputs unclustered tracks.

Then maybe they should go in another PR with their own explanation rather than here?

andread3vita:
> Then maybe they should go in another PR with their own explanation rather than here?

I agree, but wouldn't changing the title and description of this PR have the same result? I know that's probably not the best approach, so if you prefer, I can revert these commits and open a new PR once this one has been approved. Or, we can close this PR and open a new one with all this fixes.

mahmoudali2:
> Hi! With the latest commits, the issue should now be resolved without relying on a try-catch block. @mahmoudali2, could you please test if it works for you?
> 
> This PR also includes a few other commits with fixes I spotted last week. Additionally, I've made a change to the track finder so that it no longer outputs unclustered tracks.

The crash is still reproducible after applying the fixes. You also have the file that triggers it, have you tried running it? In my case, after removing the try-catch block, the crash occurs again.

andread3vita:
> > Hi! With the latest commits, the issue should now be resolved without relying on a try-catch block. @mahmoudali2, could you please test if it works for you?
> > This PR also includes a few other commits with fixes I spotted last week. Additionally, I've made a change to the track finder so that it no longer outputs unclustered tracks.
> 
> The crash is still reproducible after applying the fixes. You also have the file that triggers it, have you tried running it? In my case, after removing the try-catch block, the crash occurs again.

Hi! Yes, I have the file to reproduce the error and it is not crashing anymore. I am using this PR to test it and the file I am using is here: https://cernbox.cern.ch/files/spaces/eos/user/a/adevita/public/z2mumu_reco.edm4hep.root

andread3vita:
@jmcarcell @mahmoudali2 I think we can close this PR and open a new one where we can describe in details all the modifications. What do you think?

andread3vita:
I have opened a [new PR](https://github.com/key4hep/k4RecTracker/pull/88) addressing all the issues. @mahmoudali2 could you test it? I would close this one in the meantime

mahmoudali2:
This has been replaced by #88

PR #86, opened 2026-07-02
Fix TracksFromGenParticles - build tracks only for particles that reach the calorimeter

SanghyunKo:
BEGINRELEASENOTES
- Fix TracksFromGenParticles - build tracks only for particles that reach the calorimeter

ENDRELEASENOTES

Currently a track is built for every charged MCParticle with a matched hit, which gives unphysical tracks: unstable baryons (Σ, Ξ, Ω, D…) that decay in the tracker, and both the parent and daughter of an in-flight decay (e.g. π/K → μ), so the same trajectory is double-counted.

Fix: skip particles with `isDecayedInTracker()` (new `OnlyCaloReachingParticles` property, default true). This drops the ghosts and in-flight-decay parents while keeping the calo-reaching daughter. Set the property false for the old behaviour.

Also skip charged particles with pt = 0 (exactly along the beam): their helix is singular (omega, tanLambda, Z0 → non-finite) and crashes the downstream Pandora track creation.

### DIFF
--- a/Tracking/components/TracksFromGenParticles.cpp
+++ b/Tracking/components/TracksFromGenParticles.cpp
@@ -191,8 +191,21 @@ struct TracksFromGenParticles final
       if (genParticle.getCharge() == 0)
         continue;
 
+      // Build tracks only for charged particles that reach the calorimeter face.
+      // isDecayedInTracker() (endpoint inside the tracker) rejects the particles
+      // whose helix extrapolation to the calo is a ghost: in-flight-decay parents,
+      // unstable baryons and low-pT curlers that stop in the tracker.  Real
+      // calo-reaching daughters/secondaries are kept regardless of generatorStatus.
+      if (m_onlyCaloReachingParticles && genParticle.isDecayedInTracker())
+        continue;
+
       // skip low momentum particles that cannot be reconstructed in tracker
       if (pmag < m_minParticleMomentum)
+        continue;
+
+      // A charged particle with exactly zero transverse momentum travels along the beam axis;
+      // its helix is singular (omega = qB/pt, tanLambda = pz/pt, Z0 all diverge to non-finite).  Skip it.
+      if (p.x * p.x + p.y * p.y < std::numeric_limits<double>::epsilon())
         continue;
 
       // build an helix out of MCParticle properties and B field
@@ -431,6 +444,14 @@ private:
   Gaudi::Property<float> m_minParticleMomentum{
       this, "MinimumParticleMomentum", 0.010,
       "Keep only particles with momentum (in GeV) greater than MinimumParticleMomentum"};
+
+  /// Build tracks only for charged particles reaching the calorimeter, i.e. NOT
+  /// isDecayedInTracker() (rejects in-flight-decay parents, unstable baryons and
+  /// low-pT curlers whose helix extrapolation to the calo is a ghost).  When
+  /// false a track is built for every charged particle with a matched SimTrackerHit.
+  Gaudi::Property<bool> m_onlyCaloReachingParticles{this, "OnlyCaloReachingParticles", true,
+                                                    "Build tracks only for charged particles reaching the calorimeter "
+                                                    "(not isDecayedInTracker)"};
 
   /// Configurable property listing the systemIDs of the variuos tracker subdetectors
   Gaudi::Property<std::vector<int>> m_trackerIDs{this, "TrackerIDs", {}, "System IDs of tracking subdetectors"};
--- a/Tracking/components/TracksFromGenParticles.cpp
+++ b/Tracking/components/TracksFromGenParticles.cpp
@@ -191,11 +191,8 @@ struct TracksFromGenParticles final
       if (genParticle.getCharge() == 0)
         continue;
 
-      // Build tracks only for charged particles that reach the calorimeter face.
-      // isDecayedInTracker() (endpoint inside the tracker) rejects the particles
-      // whose helix extrapolation to the calo is a ghost: in-flight-decay parents,
-      // unstable baryons and low-pT curlers that stop in the tracker.  Real
-      // calo-reaching daughters/secondaries are kept regardless of generatorStatus.
+      // Build tracks only for charged particles that reach the calorimeter face
+      // if m_onlyCaloReachingParticles is set to True
       if (m_onlyCaloReachingParticles && genParticle.isDecayedInTracker())
         continue;
 


SanghyunKo:
Release CI is expected to fail without key4hep/k4geo#581.

For nightly CI, it seems that one of the tests is using too much memory (nothing relevant to the PR):
```
real	4m3.444s
user	10m2.917s
sys	1m23.002s
Cacheable calls:    60 /  60 (100.0%)
  Hits:              0 /  60 ( 0.00%)
    Direct:          0
    Preprocessed:    0
  Misses:           60 /  60 (100.0%)
Local storage:
  Cache size (GB): 0.5 / 0.5 (103.1%)
  Cleanups:         50
  Hits:              0 /  60 ( 0.00%)
  Misses:           60 /  60 (100.0%)
Test project /k4rectracker/build
    Start 3: test_runDCHdigiV1
    Start 8: test_TrackFinder
    Start 1: test_DCHsimpleDigitizer
    Start 2: test_DCHsimpleDigitizerExtendedEdm
1/9 Test #3: test_runDCHdigiV1 ....................   Passed   89.17 sec
    Start 4: test_runDCHdigiV2
2/9 Test #4: test_runDCHdigiV2 ....................   Passed    2.91 sec
    Start 5: test_runVTXdigitizer
3/9 Test #1: test_DCHsimpleDigitizer ..............   Passed  180.69 sec
    Start 6: test_VTXdigitizerDetailed
4/9 Test #2: test_DCHsimpleDigitizerExtendedEdm ...   Passed  182.12 sec
    Start 7: test_TracksFromGenParticles
5/9 Test #6: test_VTXdigitizerDetailed ............   Passed   61.42 sec
/home/runner/work/_temp/d69dfa95-b515-44b1-8a9a-d59ba5052669.sh: line 115:  2431 Killed                  docker exec container /bin/bash -c "/mount.sh && /${name}/script_container.sh"
Error: Process completed with exit code 137.
```

Tracking/components/TracksFromGenParticles.cpp
```diff
@@ -191,10 +191,23 @@ struct TracksFromGenParticles final
...
+      if (m_onlyCaloReachingParticles && genParticle.isDecayedInTracker())
+        continue;
+
       // skip low momentum particles that cannot be reconstructed in tracker
       if (pmag < m_minParticleMomentum)
         continue;
 
+      // A charged particle with exactly zero transverse momentum travels along the beam axis;
+      // its helix is singular (omega = qB/pt, tanLambda = pz/pt, Z0 all diverge to non-finite).  Skip it.
+      if (p.x * p.x + p.y * p.y < std::numeric_limits<double>::epsilon())
```
jbeirer, line 205:
```suggestion
      const auto pt2 = p.x * p.x + p.y * p.y;
      if (pt2 < std::numeric_limits<double>::epsilon())
```

jmcarcell: If one really wants you can do `(if auto pt2 = p.x * p.x + p.y * p.y < ...)` for clarity but I would say it's short enough and `pt2` is not used anywhere else.

Tracking/components/TracksFromGenParticles.cpp
```diff
@@ -191,10 +191,23 @@ struct TracksFromGenParticles final
       if (genParticle.getCharge() == 0)
         continue;
 
+      // Build tracks only for charged particles that reach the calorimeter face.
```
jmcarcell: Can the comments be tuned down? This is almost a duplicate of the one for the new property

SanghyunKo: Done

jmcarcell:
Looks good to me but it changes the current behaviour. @andread3vita any comments?

andread3vita:
@SanghyunKo, did you notice any difference in the particles reaching the calorimeter with and without these modifications?

SanghyunKo:
@andread3vita yes, this will kill tracks built from particles that decayed inside the tracker. For example, for π/K->μ+ν in-flight decay, this will prevent building tracks for both pi/K and μ and subsequent track double-counting. Also, currently `TracksFromGenParticles` contains tracks built from unstable baryons like Σ, Ξ, Ω, D, ... which is simply not physical at all.

andread3vita:
> @andread3vita yes, this will kill tracks built from particles that decayed inside the tracker. For example, for π/K->μ+ν in-flight decay, this will prevent building tracks for both pi/K and μ and subsequent track double-counting. Also, currently `TracksFromGenParticles` contains tracks built from unstable baryons like Σ, Ξ, Ω, D, ... which is simply not physical at all.

But you're doing that because the propagation to the calorimeter fails for those tracks, right? Wouldn't it be better to keep the tracks and simply skip the propagation for them, leaving them without a `TrackState` at the calorimeter?

SanghyunKo:
No it's not about the extrapolation. The point is that the downstream algorithms relies on the emitted track collection, and I cannot filter/cherry-pick tracks within the same collection. If we import the collection in Gaudi, we have to deal with the all components inside the collection. So it's either:
1. Manually filter out unphysical tracks in ALL downtream algorithms that use this track collection (i.e. if we have N consumers downstream then we spam the filter N times).
2. Create another Gaudi functional that simply filters out unphysical tracks and save remainders to a separated track collection (2 Gaudi functionals + 2 collections).
3. or simply add an option in this algorithm - this PR.

which I think option 1 & 2 simply make no sense. There is also no point of creating tracks out of exotic baryons, which should never exist in the first place.

andread3vita:
> No it's not about the extrapolation. The point is that the downstream algorithms relies on the emitted track collection, and I cannot filter/cherry-pick tracks within the same collection. If we import the collection in Gaudi, we have to deal with the all components inside the collection. So it's either:
> 
>     1. Manually filter out unphysical tracks in ALL downtream algorithms that use this track collection (i.e. if we have N consumers downstream then we spam the filter N times).
> 
>     2. Create another Gaudi functional that simply filters out unphysical tracks and save remainders to a separated track collection (2 Gaudi functionals + 2 collections).
> 
>     3. or simply add an option in this algorithm - this PR.
> 
> 
> which I think option 1 & 2 simply make no sense. There is also no point of creating tracks out of exotic baryons, which should never exist in the first place.

Then it is fine for me if we go for the third option

jmcarcell:
> Then it is fine for me if we go for the third option

Merging then, note this changes the behaviour since the option is enabled by default.

PR #87, opened 2026-07-08
Add track merging algorithm based on track parameter compatibility

Victor-Schwan:
BEGINRELEASENOTES
- Adds a new `TrackMerger` Gaudi transformer that merges tracks from two input collections based on compatible track-state parameters at their adjoining hit
- While generic and applicable to merging any two track collections, it was designed to combine silicon-tracker tracks with TPC tracks reconstructed by `Clupatra` for ILD
- The exact matching criterion in track-parameter space is not yet optimized and may need tuning
- Includes self-contained CTest `test_TrackMerger*` that generates synthetic matching/non-matching track pairs and verifies correct merging behavior

ENDRELEASENOTES

### DIFF
--- a/Tracking/CMakeLists.txt
+++ b/Tracking/CMakeLists.txt
@@ -92,4 +92,9 @@ set_test_env(${test_name})
 set_tests_properties(${test_name} PROPERTIES DEPENDS test_TrackFinder)
 set_tests_properties(${test_name} PROPERTIES WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}/Tracking/test/testTrackFitter")
 
+SET(test_name "test_TrackMerger")
+ADD_TEST(NAME ${test_name} COMMAND python3 test_track_merger.py)
+set_test_env(${test_name})
+set_tests_properties(${test_name} PROPERTIES WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}/Tracking/test/testTrackMerger")
+
 #endif()
--- a/Tracking/CMakeLists.txt
+++ b/Tracking/CMakeLists.txt
@@ -95,12 +95,21 @@ set_tests_properties(${test_name} PROPERTIES WORKING_DIRECTORY "${CMAKE_SOURCE_D
 SET(test_name "test_TrackMerger")
 SET(test_dir "${CMAKE_SOURCE_DIR}/Tracking/test/testTrackMerger")
 
+ADD_TEST(NAME ${test_name}_setup
+  COMMAND python3 test_track_merger.py setup --input input.root)
+set_test_env(${test_name}_setup)
+set_tests_properties(${test_name}_setup PROPERTIES
+  WORKING_DIRECTORY "${test_dir}"
+  FIXTURES_SETUP ${test_name}_input_fixture
+)
+
 ADD_TEST(NAME ${test_name}_run
-  COMMAND python3 test_track_merger.py run --input input.root --output output.root)
+  COMMAND k4run test_track_merger_steer.py --input input.root --output output.root)
 set_test_env(${test_name}_run)
 set_tests_properties(${test_name}_run PROPERTIES
   WORKING_DIRECTORY "${test_dir}"
-  FIXTURES_SETUP ${test_name}_fixture
+  FIXTURES_REQUIRED ${test_name}_input_fixture
+  FIXTURES_SETUP ${test_name}_output_fixture
 )
 
 ADD_TEST(NAME ${test_name}_check
@@ -108,7 +117,6 @@ ADD_TEST(NAME ${test_name}_check
 set_test_env(${test_name}_check)
 set_tests_properties(${test_name}_check PROPERTIES
   WORKING_DIRECTORY "${test_dir}"
-  FIXTURES_REQUIRED ${test_name}_fixture
+  FIXTURES_REQUIRED ${test_name}_output_fixture
 )
-
 #endif()
--- a/Tracking/CMakeLists.txt
+++ b/Tracking/CMakeLists.txt
@@ -111,11 +111,4 @@ set_tests_properties(${test_name}_check PROPERTIES
   FIXTURES_REQUIRED ${test_name}_fixture
 )
 
-ADD_TEST(NAME ${test_name}_cleanup
-  COMMAND ${CMAKE_COMMAND} -E rm -f input.root output.root)
-set_tests_properties(${test_name}_cleanup PROPERTIES
-  WORKING_DIRECTORY "${test_dir}"
-  FIXTURES_CLEANUP ${test_name}_fixture
-)
-
 #endif()
--- a/Tracking/CMakeLists.txt
+++ b/Tracking/CMakeLists.txt
@@ -93,8 +93,29 @@ set_tests_properties(${test_name} PROPERTIES DEPENDS test_TrackFinder)
 set_tests_properties(${test_name} PROPERTIES WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}/Tracking/test/testTrackFitter")
 
 SET(test_name "test_TrackMerger")
-ADD_TEST(NAME ${test_name} COMMAND python3 test_track_merger.py)
-set_test_env(${test_name})
-set_tests_properties(${test_name} PROPERTIES WORKING_DIRECTORY "${CMAKE_SOURCE_DIR}/Tracking/test/testTrackMerger")
+SET(test_dir "${CMAKE_SOURCE_DIR}/Tracking/test/testTrackMerger")
+
+ADD_TEST(NAME ${test_name}_run
+  COMMAND python3 test_track_merger.py run --input input.root --output output.root)
+set_test_env(${test_name}_run)
+set_tests_properties(${test_name}_run PROPERTIES
+  WORKING_DIRECTORY "${test_dir}"
+  FIXTURES_SETUP ${test_name}_fixture
+)
+
+ADD_TEST(NAME ${test_name}_check
+  COMMAND python3 test_track_merger.py check --output output.root)
+set_test_env(${test_name}_check)
+set_tests_properties(${test_name}_check PROPERTIES
+  WORKING_DIRECTORY "${test_dir}"
+  FIXTURES_REQUIRED ${test_name}_fixture
+)
+
+ADD_TEST(NAME ${test_name}_cleanup
+  COMMAND ${CMAKE_COMMAND} -E rm -f input.root output.root)
+set_tests_properties(${test_name}_cleanup PROPERTIES
+  WORKING_DIRECTORY "${test_dir}"
+  FIXTURES_CLEANUP ${test_name}_fixture
+)
 
 #endif()
--- a/Tracking/components/TrackMerger.cpp
+++ b/Tracking/components/TrackMerger.cpp
@@ -1,0 +1,150 @@
+/*
+ * Copyright (c) 2020-2026 Key4hep-Project.
+ *
+ * This file is part of Key4hep.
+ * See https://key4hep.github.io/key4hep-doc/ for further info.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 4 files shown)


Tracking/components/TrackMerger.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+    debug() << fmt::format("    Comparing Loc {} vs {}: d0_diff={:.4f}, z0_diff={:.4f} -> Match: {}", loc1, loc2,
+                           d0_diff, z0_diff, match)
+            << endmsg;
+
+    return match;
+  }
+
+  std::optional<const TS> getTrackState(const Track& track, const int loc) const {
+    const auto& trackStates = track.getTrackStates();
```
jmcarcell: You can pass the location to `getTrackStates`, making this function not be needed: https://github.com/key4hep/EDM4hep/blob/main/edm4hep.yaml#L545

Tracking/components/TrackMerger.cpp
```diff
@@ -0,0 +1,150 @@
...
+    if (!ts1.has_value() || !ts2.has_value()) {
+      // It's common for some tracks to lack specific states; verbose instead of debug to avoid spam
+      warning() << fmt::format("    [SKIP] Missing requested states (Loc1: {}, Loc2: {})", loc1, loc2) << endmsg;
+      return false;
+    }
+
+    // Define matching criteria based on d0 and z0 differences
+    const float d0_diff = std::abs(ts1->D0 - ts2->D0);
+    const float z0_diff = std::abs(ts1->Z0 - ts2->Z0);
+    const bool match = (d0_diff <= 0.5f && z0_diff <= 2.5f);
```
jmcarcell: Hardcoded values of the thresholds 😕

Tracking/components/TrackMerger.cpp
```diff
@@ -0,0 +1,150 @@
...
+using TrackColl = edm4hep::TrackCollection;
+using Track = edm4hep::Track;
+using TP = edm4hep::TrackParams;
+using TS = edm4hep::TrackState;
+
+struct TrackMerger final : k4FWCore::Transformer<TrackColl(const TrackColl&, const TrackColl&)> {
+  TrackMerger(const std::string& name, ISvcLocator* svcLoc)
+      : Transformer(name, svcLoc,
+                    {
+                        KeyValue("InputSiTracks", {"SiTracksCT"}),
```
jmcarcell: I think there should be a different name for the input collections, otherwise they don't make any sense outside of the iLCSoft context.

Tracking/components/TrackMerger.cpp
Victor-Schwan: I have wrapped the edm4hep-native function to preserve the warning message

Tracking/components/TrackMerger.cpp
Victor-Schwan: Good catch! Now, the tolerance for every track param difference is configurable in the steering file

Tracking/components/TrackMerger.cpp
Victor-Schwan: Fair point. I changed it to `Inner` or `Outer` track. The naming refers to whether the last hit track state (inner) or first hit track state (outer) is used.

Tracking/test/testTrackMerger/test_track_merger.py
```diff
@@ -0,0 +1,240 @@
+"""
```
jmcarcell, line 1: Why not call k4run from `CMakeLists.txt` like all the tests everywhere else do? Also you can have a steering file and modify parameters.

jmcarcell:
It seems I forgot to write this: the results of this algorithm depend on the ordering of the input tracks, which is not nice. Also, this is only useful for something downstream that wants the hits but instead if you return a link between each pair of tracks then someone could have available all the information and you don't have to create new Track objects.

Tracking/test/testTrackMerger/test_track_merger.py
Victor-Schwan: My idea was to have a self-contained script.
1. All data of my test is in a temporary directory and after execution all files are removed and the git repository is in a clean state. The other tests do not remove their files and litter my repository.
2. The script creates its own input and steering files -> no need to download input files or steering files as it is in the other tests
3. The `test_trackFitter.sh` basically only calls the python script. I am not aware of a significant advantage that the wrapping offers
4. You can modify parameters in my python script as well. Parameters like the tolerances for merging and collection names, are located at the top of the script for convenient adaptation

Of course, one can split the different steps in this python script over several files but why?

Victor-Schwan:
1) On the dependence of track ordering:
This is only the case if the `greedy` option is activated. I suggest that we change the default to deactivate this option. Then, the default behavior does not depend on the ordering. We can provide a warning and comment that this option introduces a dependence on track ordering. I discussed with Thomas that we would like to study this option. Hence, I would like to keep it in the code. Is that fine?
2) On providing the links:
What about adding an option which writes out the links? If someone is interested in the links, they can activate this option.

Tracking/test/testTrackMerger/test_track_merger.py
jmcarcell: You can just add the files to `.gitignore` and you will never notice the files from git. Indeed there is no reason to have `test_trackFitter.sh` instead of calling directly the k4run command; I just see it now. Creating the input file is fine and for steering files we already have a mechanism to change parameters instead of having to generate it; this also allows it to be run independently instead of having to find out the python script, then the generation, then which command runs, it's all much more complex than needed for a simple test.

jmcarcell:
Ok for 1. Also `greedy` typically means taking the best option at each step (https://en.wikipedia.org/wiki/Greedy_algorithm), which would be to find the "best" track for each input track and using that one even if globally it's not the best matching. Slightly different from first that passes but I'm not sure which word reflects that better.
2. Well actualy downstream algorithms would need to work with links instead of Tracks so it's also not the best setup so I guess Tracks is fine.

Victor-Schwan:
IIUC points 1) and 2) are solved. If not, please let me know

Tracking/test/testTrackMerger/test_track_merger.py
Victor-Schwan: Ok. Do you think there are more adaptations necessary than promoting the steering script to a standalone file? This results in outsourcing step 2 of the python script but the rest remains as it is?

Victor-Schwan: Compared to the other tests, the track merger test has to check whether one track pair was merged and the other pair not. AFAIK the other tests pass as long as the processors were executed without any problem. This verification step, the input file generation and the execution would remain in the python script.

jmcarcell: > Ok. Do you think there are more adaptations necessary than promoting the steering script to a standalone file? This results in outsourcing step 2 of the python script but the rest remains as it is?

That should be enough, that almost forces you to have two CMake tests, one for running and another one for checking, you can make the check depend on the running passing with CMake fixtures. This is done in k4FWCore if you want to see some examples.


> Compared to the other tests, the track merger test has to check whether one track pair was merged and the other pair not. AFAIK the other tests pass as long as the processors were executed without any problem. This verification step, the input file generation and the execution would remain in the python script.

Ideally they should check their output too, they also pass if they process 0 events

jmcarcell:
> IIUC points 1) and 2) are solved. If not, please let me know

I don't see any new commits

Victor-Schwan:
> I don't see any new commits

That is because there are no new commits. I understood your replies as follows:

1) > Ok for 1. [...] Slightly different from first that passes but I'm not sure which word reflects that better.

Since neither of us has a better idea for a more precise wording, I assumed we are fine with leaving it as it is

2) > [...] so I guess Tracks is fine

Currently, the code writes out tracks with the merged hit collections. Previously, you had thought about writing out links for a downstream algo but you changed your mind and thought that tracks (the way it is currently implemented) is fine. Therefore I assumed you were fine with leaving it as it is.

---

Hence, I concluded there is no need for any changes/new commits. Please let me know what I have misunderstood :)

Tracking/test/testTrackMerger/test_track_merger.py
Victor-Schwan: Should be solved by the last commit. Based on your suggestion, I've created two tests that depend on each other via fixtures, and a cleanup "test".

jmcarcell:
By the way, what do we do with https://github.com/key4hep/k4RecTracker/pull/72?

Tracking/CMakeLists.txt
```diff
@@ -92,4 +92,30 @@ set_test_env(${test_name})
...
+ADD_TEST(NAME ${test_name}_check
+  COMMAND python3 test_track_merger.py check --output output.root)
+set_test_env(${test_name}_check)
+set_tests_properties(${test_name}_check PROPERTIES
+  WORKING_DIRECTORY "${test_dir}"
+  FIXTURES_REQUIRED ${test_name}_fixture
+)
+
+ADD_TEST(NAME ${test_name}_cleanup
+  COMMAND ${CMAKE_COMMAND} -E rm -f input.root output.root)
```
jmcarcell: What is so bad about having the root files around? If added to .gitignore they won't be seen. The working directory can be inside the build directory if you don't want them to be in the source directory.

Victor-Schwan: If a test is rerun, isn't there a possibility that output files from a previous run interfere with the rerun? Maybe existing output from the run step causes this step to be skipped? The other tests also remove output from previous runs (see e.g. https://github.com/key4hep/k4RecTracker/blob/c4054cc5257fac15760e850d7b05b5a6bfdf0ccd/Tracking/test/testTrackFinder/test_trackFinder.sh#L4-L5)

Otherwise, I am also fine with your approach

Victor-Schwan:
> By the way, what do we do with #72?

Good question. @tmadlener  was pushing for merging this, so I would like to discuss that with him when he is back which will be soon. This will certainly not affect this PR. The question will be whether we add #72 in addition, in a modified version or not at all.

Tracking/CMakeLists.txt
jmcarcell: Not with CMake's fixtures because if the setup test fails then the actual test doesn't run (unlike with `DEPENDS`, where it always runs, just after). And also since the runs from CI run in a clean state, it's quite hard for that situation to happen; if you check in other repositories you'll see we don't worry about deleting output files in tests.

Victor-Schwan: That makes sense for the CI, but how does it work when running the tests repeatedly locally during development? Without a cleanup, there is a risk of testing against stale data. Eg, if a step fails silently or skips writing output, a downstream test might falsely pass by reading the `output.root` left over from a previous run.

In any case, I have removed the cleanup test in the latest commit as requested.

jmcarcell: It's the same, if it crashes then downstream tests shouldn't run and that won't happen with fixtures, or what is the setup you are thinking about? If it's you who is running them manually well then you should use `&&` or `set -e` in bash, ideally you want to stop as soon as the failure happens, not continue to downstream and let something complain because the file is not there.

Tracking/test/testTrackMerger/test_track_merger.py
jmcarcell: But this does not call k4run from CMakeLists which was my original comment, isn't that simpler than calling subprocess? Then the logs are stored in a file and you don't need to do 
```
    print(result.stdout[-3000:])  # tail to keep output readable
```
What if the failure happens in characters that are removed? Now you have to run it, ah where is the command... inside the script, where I think no one would expect it. Also if you ever want to use this script outside of this repo there are many changes that have to be done because of the complex setup, instead of just taking input and output files.

Victor-Schwan: ok, the truncated-`stdout` issue alone could've been fixed by just dropping `capture_output=True` (then everything flows through to CTest's log), but I agree the `k4run` command being hidden inside the script makes reproducing failures annoying. I've now split it into three CTest steps chained via fixtures:
1) python setup (writes input file)
2) k4run called directly from CMakeLists
3) python check

The steering file only takes `--input/--output`, so it's also reusable outside the repo. If you'd rather keep it consistent with the other tests, I could alternatively wrap setup + k4run in a small bash script but I do not see an advantage of that.

Are you fine with this update?

jmcarcell:
Unrelated CI failures (related to https://github.com/key4hep/key4hep-spack/pull/898)

PR #88, opened 2026-07-28
Tracking: fix post-fit TrackState, reset TGeoManager before each fit, correct the Jacobian, remove noise tracks from GGTF output and suppress error messages with debugLvl=INFO

andread3vita:
BEGINRELEASENOTES

- Corrected the units used in the Jacobian to ensure consistency between GENFIT and EDM4hep.
- Fixed broken associations between hits and tracks introduced during the filtering stage.
- Refactored the track-finding algorithm to discard tracks classified as noise. Only successfully reconstructed tracks are now written to the output file.
- Reset the global TGeoManager navigator to a canonical state before each fit. The navigator state is process-global and is modified by material lookups during track extrapolation. Without this reset, the material assigned to points located on geometry boundaries may depend on the previously fitted track, potentially from an earlier event, making the fit results dependent on processing order.
- Updated the TrackState objects at the first and last hits to use the measured hit positions as reference points when computing d0 and z0.
- Added exception handling around getFittedState after the GENFIT fit. A failure to retrieve the fitted state for a single track no longer terminates the entire process.
- Added the GenfitDebugLevel.h utility, including the ScopedFitDiagnostics RAII guard, to temporarily suppress GENFIT error and debug output during track fitting. The utility also filters the expected ROOT error emitted by TDecompChol::Decompose for non-positive-definite matrices, while forwarding all other ROOT diagnostics and automatically restoring the original stream buffers and error handler when leaving the fitting scope.

ENDRELEASENOTES

Hello!
I am opening this PR to address some bugs reported over the past few weeks. It also introduces a utility to suppress diagnostic messages generated by failed fit attempts. These messages remain available when debugLvl is set to DEBUG or VERBOSE.

Cheers,
Andrea

### DIFF
--- a/Tracking/components/GGTFTrackFinder.cpp
+++ b/Tracking/components/GGTFTrackFinder.cpp
@@ -18,46 +18,31 @@
  */
 
 // Standard Library
-#include <algorithm>
+#include <atomic>
 #include <cmath>
-#include <cstdlib>
-#include <filesystem>
-#include <fstream>
-#include <iostream>
-#include <iterator>
-#include <map>
+#include <cstdint>
+#include <limits>
 #include <memory>
-#include <numeric>
-#include <queue>
-#include <random>
-#include <sstream>
+#include <stdexcept>
 #include <string>
-#include <typeinfo>
+#include <tuple>
 #include <vector>
 
-// ONNX & Torch
+// ONNX Runtime and Torch
 #include "onnxruntime_cxx_api.h"
-#include "onnxruntime_run_options_config_keys.h"
-#include <ATen/ATen.h>
 #include <torch/torch.h>
 
 // ROOT
 #include "TVector3.h"
 
 // Gaudi
-#include "Gaudi/Algorithm.h"
 #include "Gaudi/Property.h"
 
-// k4FWCore / k4Interface
-#include "k4FWCore/DataHandle.h"
+// k4FWCore
 #include "k4FWCore/Transformer.h"
-#include "k4Interface/IGeoSvc.h"
-#include "k4Interface/IUniqueIDGenSvc.h"
 
 // EDM4HEP
-#include "edm4hep/MCParticleCollection.h"
 #include "edm4hep/SenseWireHitCollection.h"
-#include "edm4hep/SimTrackerHitCollection.h"
 #include "edm4hep/TrackCollection.h"
 #include "edm4hep/TrackerHitPlaneCollection.h"
 
@@ -67,9 +52,10 @@
  *
  * Gaudi MultiTransformer that generates a Track collection by analyzing the digitalized hits through the
  * GGTFTrackFinder. The first step takes the raw hits and it returns a collection of 4-dimensional points inside an
- * embedding space. Eeach 4-dim point has 3 geometric coordinates and 1 charge, the meaning of which can be described
- * intuitively by a potential, which attracts hits belonging to the same cluster and drives away those that do not. This
- * collection of 4-dim points is analysed by a clustering step, which groups together hits belonging to the same track.
+ * embedding space. Each 4-dim point has 3 geometric coordinates and 1 charge, the meaning of which can be described
+ * intuitively by a potential, which attracts hits belonging to the same cluster and drives away those that do not.
+ * This collection of 4-dim points is analysed by a clustering step, which groups together hits belonging to the
+ * same track.
  *
  *  input:
  *    - digitalized hits from DC (global coordinates) : edm4hep::SenseWireHitCollection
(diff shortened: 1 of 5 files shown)


andread3vita:
FYI: @mahmoudali2

mahmoudali2:
> FYI: @mahmoudali2

It works without crashes, you can go a head with the merge.

Tracking/components/GGTFTrackFinder.cpp
```diff
@@ -83,356 +69,365 @@
...
+      }
 
-    /// Processing hits from the Silicon Detectors
-    std::vector<int64_t> listHitTypePlanar;
-    int planarHitIndex = 0;
-    std::vector<int64_t> listPlanarHitIndices;
-    int planarHitCollectionIndex = 0;
-    for (const auto& inputHitCollection : inputPlanarHitCollections) {
+      for (std::size_t hitIndex = 0; hitIndex < collection->size(); ++hitIndex) {
+        if (batch.nHits > static_cast<std::size_t>(kMaxHits)) {
```
jmcarcell, line 226:
```suggestion
        if (batch.nHits >= static_cast<std::size_t>(kMaxHits)) {
```
Currently it accepts also `kMaxHits+1`.

Tracking/include/genfit_interfaces/GenfitDebugLevel.h
```diff
@@ -0,0 +1,73 @@
+#ifndef GENFIT_DEBUG_LEVEL_H
+#define GENFIT_DEBUG_LEVEL_H
+
+#include <cstring>
+#include <streambuf>
+
+#include <IO.h>
+#include <TError.h>
+#include <TGeoManager.h>
+
```
jmcarcell, line 10:
```suggestion
```
ROOT is not used in this file. On the other side genfit (or the genfit namespace) is but there is not any include for genfit.

Tracking/src/genfit_interfaces/GenfitTrack.cpp
```diff
@@ -904,10 +908,15 @@ bool GenfitTrack::Fit(std::string FitterType = "DAF", int debug_lvl = 0, std::op
...
-    debug_lvl_fit = 0;
-  genfitFitter->setDebugLvl(debug_lvl_fit);
+  genfitFitter->setDebugLvl(showFitterDiagnostics ? 1 : 0);
+
+  // Reset the global TGeoManager navigator to a canonical state before fitting.
+  // The navigator (current node, point, direction, safety/step caches) is process-global
+  // and is mutated by every material lookup during extrapolation. Without this reset,
+  // the material resolved for on-boundary points can depend on the track fitted
+  // previously (even in an earlier event), making fit results order-dependent.
+  gGeoManager->CdTop();
```
jmcarcell, line 918: It would be good to make this algorithm not-reentrant like in https://github.com/key4hep/k4FWCore/blob/main/k4FWCore/components/Writer.cpp#L50 since these changes of global state plus the diagnostic guards are not going to work well in multithreading (probably no one has tried to run it multithreaded?).

jmcarcell (commented):
This PR is very hard to review with many different changes, it would benefit from splitting the smaller changes from the biggest refactors.

Tracking/src/genfit_interfaces/GenfitTrack.cpp
```diff
@@ -117,9 +120,18 @@ void GenfitTrack::OrderHits(const edm4hep::Track& track, bool skipTrackOrdering)
...
+    const auto& hit = hits[idx];
+    auto pos = hit.getPosition();
+    m_edm4hepTrack.addToTrackerHits(hit);
+
+    if (first) {
+      m_FirstHit_referencePoint = TVector3(pos.x * dd4hep::mm, pos.y * dd4hep::mm, pos.z * dd4hep::mm);
+      first = false;
+    }
+
+    m_LastHit_referencePoint = TVector3(pos.x * dd4hep::mm, pos.y * dd4hep::mm, pos.z * dd4hep::mm);
```
jmcarcell, line 134: These lines can never be reached if `m_skipTrackOrdering` is true; I think that is not supposed to happen.

SanghyunKo:
I'm not sure whether this is still active, but I drop my (macroscopic) comments nevertheless:

As @jmcarcell already mentioned, this is a bit hard to review as one piece. I'd propose to split the PR into 3:

1. `GGTFTrackFinder.cpp` on its own — separate component, no overlap with the GenFit fitter, and it validates on track multiplicity rather than fit quality.
2. Trivial GenFit fixes that need no validation: the delete `m_genfitTrackRep` double-free, the null-initialised `m_wire_info`/`m_dc_decoder`, and the `Init_position` unit conversion that was missing in `FindBestHypothesis` on main. Worth putting the comment reflowing and log-level changes here too. This one can merge quickly.
3. The substantive changes: Jacobian/units, the Sigma_d0/z0 convention, the TrackState reference points, the `TGeoManager` reset and the diagnostics guard — all sharing one validation, resolution and pull distributions before and after.

One request either way: `281be92e` currently bundles four independent fixes, so please use one commit per concern. Three fixes also deserve a line in the release notes not the one (the double-free, the missing `Init_position` conversion, and the Sigma_d0/z0 convention change).

andread3vita:
I close this PR because I am going to create three more specific PRs

jmcarcell:
Please address the comments I made above if you didn't in the new PRs

Tracking/src/genfit_interfaces/GenfitTrack.cpp
andread3vita: Done

Tracking/src/genfit_interfaces/GenfitTrack.cpp
andread3vita: I will open a new PR with the fix

Tracking/components/GGTFTrackFinder.cpp
andread3vita: Done

Tracking/include/genfit_interfaces/GenfitDebugLevel.h
andread3vita: I will open a PR with the fixed version of the debug handler

PR #92, opened 2026-08-07
TrackdNdxDelphesBased Fix

s6anloes:
BEGINRELEASENOTES
- Adapted and changed logic of accepted $\beta\gamma$ range in `TrackdNdxDelphesBased`

ENDRELEASENOTES

I recently noticed that the delphes accepted range for $\beta\gamma$ was lowered (whether intentionally or not, see https://github.com/delphes/delphes/issues/199) to $10000$.
Thus, in the [`run_digi_reco.py`](https://github.com/HEP-FCC/FCC-config/blob/main/FCCee/FullSim/IDEA/IDEA_o1_v03/run_digi_reco.py) script in FCC-Config for `IDEA`, the 10 GeV electrons would silently be out of the range and produce `nclusters=0`.

This PR aligns the upper limit of accepted $\beta\gamma$ in `TrackdNdxDelphesBased` with the range from `delphes/TrkUtil`.
Also, the logic is slighty changed, so that if the value is above the range, it will set $\beta\gamma$ to the max value as an approximation. This way 10 GeV electrons will still produce clusters.

Cheers,
Andreas

### DIFF
--- a/Tracking/components/TrackdNdxDelphesBased.cpp
+++ b/Tracking/components/TrackdNdxDelphesBased.cpp
@@ -109,11 +109,15 @@ edm4hep::RecDqdxCollection TrackdNdxDelphesBased::operator()(const edm4hep::Trac
       double betagamma = momentum / mass;
       debug() << "MCParticle betagamma: " << betagamma << endmsg;
       // Check if betagamma is in valid range of delphes parametrisation (status: 16 June 2025)
-      if (betagamma < 0.5 || betagamma > 20000.0) {
-        warning() << "beta*gamma value outside of \"good\" range of delphes parametrisation (0.5-20000), dN/dx will be "
+      if (betagamma < 0.5 ) {
+        warning() << "beta*gamma value below \"good\" range of delphes parametrisation (0.5-10000), dN/dx will be "
                      "set to dummy value: "
                   << dummy_value << " clusters/mm" << endmsg;
         goto store_value;
+      } else if (betagamma > 10000) {
+        warning() << "beta*gamma value above \"good\" range of delphes parametrisation (0.5-10000), "
+                     "beta*gamma will be set to max value of 10000 as approximation." << endmsg;
+        betagamma = 9999.9; // 10000 is out of range already
       }
 
       // Get number of clusters per length from delphes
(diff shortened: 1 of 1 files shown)


Tracking/components/TrackdNdxDelphesBased.cpp
```diff
@@ -109,17 +109,27 @@ edm4hep::RecDqdxCollection TrackdNdxDelphesBased::operator()(const edm4hep::Trac
       double betagamma = momentum / mass;
       debug() << "MCParticle betagamma: " << betagamma << endmsg;
       // Check if betagamma is in valid range of delphes parametrisation (status: 16 June 2025)
-      if (betagamma < 0.5 || betagamma > 20000.0) {
-        warning() << "beta*gamma value outside of \"good\" range of delphes parametrisation (0.5-20000), dN/dx will be "
+      if (betagamma < 0.5) {
+        warning() << "beta*gamma value below \"good\" range of delphes parametrisation (0.5-10000), dN/dx will be "
```
jmcarcell, line 113: The comment doesn't match, it is below 0.5

jmcarcell:
Then once it is fixed in Delphes what do you plan to do? Go back to the current version?

Tracking/components/TrackdNdxDelphesBased.cpp
s6anloes: Sorry, I don't see the mismatch?
Could you clarify?

s6anloes:
> Then once it is fixed in Delphes what do you plan to do? Go back to the current version?

Well, the approximation of setting to the max value of $\beta\gamma$ is above the range can stay I think.
We can adjust the upper limit to 20000 if it is updated in delphes again, but in any case, the 20000 value was the same as the 10000 one. It was only introduced to fix the TSpline behaviour, see https://github.com/delphes/delphes/issues/157.
So I don't expect much difference in the actual cluster per length values.
I think it could stay like this, regardless of the delphes fix

Tracking/components/TrackdNdxDelphesBased.cpp
jmcarcell: I meant that it is always below 0.5 so it doesn't make sense to say it's below 0.5-10000

ueinhaus:
The fix looks good!
It's fair to set the value in case of βγ > 10000 to the max value of the delphes function, since this is only relevant for electrons and their distribution is flat there anyway. And this can indeed be expanded/adapted following changes in delphes, though we should in general aim to fully move from this parametrisation to full sim.
One minor thing: Would it make sense to change the current out-of-range warnings to info() or even debug()? In physics events there should be regularly particles outside of the given βγ range, leading potentially to many warnings with little inherent value?

jmcarcell:
@ueinhaus does it look good now?

PR #93, opened 2026-08-07
Change geometry returned by geosvc to be used as const in GenfitTrackFitter.cpp

davidlange6:
BEGINRELEASENOTES
- Change geometry returned by geosvc to be used as const in GenfitTrackFitter.cpp
ENDRELEASENOTES

motivated by eventual change suggested in key4hep/k4SimGeant4#88

### DIFF
--- a/Tracking/components/GenfitTrackFitter.cpp
+++ b/Tracking/components/GenfitTrackFitter.cpp
@@ -123,14 +123,14 @@ struct GenfitTrackFitter final
     }
 
     // Initialize the Genfit: FieldManager and MaterialEffects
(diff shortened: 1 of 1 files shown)


jmcarcell:
`detector` is already used

davidlange6:
hum, indeed. Changed to avoid a conflict there.

jmcarcell:
Ci will fail until https://github.com/key4hep/k4SimGeant4/pull/98 is merged

davidlange6:
i would not have expected that. I guess I see it tomorrow. [oh - nevermind - i got two PRs mixed up]

